PR #42, opened 2024-07-30
Fix tests and add sanitizer to workflows

### LINKED ISSUE #41: Fix failing tests
Currently the ubuntu tests are failing

stephprince:
Fix #41.

These changes remove a test that was checking the values of a timestamps dataset that was created but never written to, fix a heap-buffer-overflow error, and add an AddressSanitizer workflow to detect memory-related errors.


src/hdf5/HDF5IO.cpp
```diff
@@ -435,7 +438,11 @@ AQNWB::BaseRecordingData* HDF5IO::createArrayDataSet(const BaseDataType& type,
 
   data = std::make_unique<H5::DataSet>(
       file->createDataSet(path, H5type, dSpace, prop));
-  return new HDF5RecordingData(data.release());
+
+  std::unique_ptr<HDF5RecordingData> recordingData =
+      std::make_unique<HDF5RecordingData>(data.release());
+
+  return recordingData.release();
```
oruebel, line 445: What does `recordingData.release()` return?  I.e., I'm just wondering whether the `recordingData` pointer is being returned or whether it is being deleted here after the function returns?

stephprince: I believe it returns a raw pointer to the HDF5RecordingData object but doesn't delete the object...

In some places (e.g. in `createArrayDataSet`) I was still returning raw pointers, I was thinking maybe I should convert those to unique pointers for automatic memory management?

oruebel: I think `unique_ptr` makes sense to use in this case, since each instance of a `BaseRecordingData` should really only exist once.

oruebel: The updated code that uses ``std::make_unique<HDF5RecordingData>(std::move(data));`` looks good.

BTW, in the original setup

```
  std::unique_ptr<HDF5RecordingData> recordingData =
      std::make_unique<HDF5RecordingData>(data.release());

  return recordingData.release();
```

 it is unclear to me what the affect of making `recordingData` a `std::unique_ptr` actually was. Because the `unique_ptr` was immediately being released after creation, I think it would have been the same as just creating it as an unmangaged pointer with `new` directly 

```
HDF5RecordingData* recordingData = new HDF5RecrodingData(data.release())
return recordingData;
```

stephprince: That makes sense. FYI I just updated the main branch protections to require approval before merging

oruebel (approved):
Looks good

Issue #39, opened 2024-07-30
Add support for Single Writer Multiple Reader (SWMR) 


stephprince:
Add support for Single Writer Multiple Reader (SWMR). After we startRecording and all datasets/groups/attributes have been created, we can turn on SWMR mode to ensure file does not get corrupted.

Issue #28, opened 2024-05-23
Add ability to stop and restart recording to same file


stephprince:
If a user temporarily stops a data stream and then restarts it during the same recording session, they may want to continue adding data to the same datasets. We should add a flag to detect whether this is a new file with existing datasets or whether new datasets like `ElectricalSeries` should be created.

stephprince:
We may need to have a difference between pausing a recording vs. stopping a recording. During start/pause/stop, SWMR mode will need to be turned on/off: https://github.com/NeurodataWithoutBorders/aqnwb/issues/39

PR #45, opened 2024-08-01
Add swmr mode

stephprince:
Fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/39.

These changes turn on SWMR mode as part of the `startRecording` process and check that public NWBFile methods that create new datasets/groups cannot be used when `isRecording` is true.

I checked that the file is accessible for reading by a separate process while open during the `uswSWMRmode` test - I'm not sure the best way to incorporate that in the tests - @oruebel any suggestions?


oruebel:
> I'm not sure the best way to incorporate that in the tests

I can see a couple of options:

1. Simply test that we can open the file again in "r" mode while the NWBFile is still open. I think this may not even need to be in a separate process. 
2. A more complete test would be to start a separate thread to read from the file while you are writing to it.
3. Have a command-line script that you are calling to read from the file

I think checking for 1 should be sufficient in the tests. SWMR is a feature of the HDF5 library (not AqNWB), so this needs to be tested in the HDF5 library.

src/hdf5/HDF5IO.cpp
```diff
@@ -377,6 +379,14 @@ Status HDF5IO::createStringDataSet(const std::string& path,
   return Status::Success;
 }
 
+Status HDF5IO::startRecording()
+{
+  if (H5Fstart_swmr_write(this->file->getId()) < 0) {
+    return Status::Failure;
+  }
```
oruebel, line 386: One question to consider is whether a user would want to be able to configure whether to enable `swmr` or not? I think at least for now, always enabling is a good idea. I think the main consequence of this is that we are enforcing a particular use pattern of:

1. setup the NWB file
2. record to the NWB file without being able to modify the structure of the file
3. pause/stop the recording process to make modifications to the file structure

I think this would in general be a best practice for data acquisition and I suspect should work for most cases. However, I think it is useful to discuss this pattern when we do outreach with the acquisition system developers to see if there may be a need to diverge from this pattern at some point. For now, I would make an issue (low priority) to explore the option to make "SWMR" mode optional by allowing the user to provide a flag to `startRecording`.

src/hdf5/HDF5IO.cpp
```diff
@@ -48,15 +49,16 @@ Status HDF5IO::open(bool newfile)
   if (opened)
     return Status::Failure;
 
-  FileAccPropList props = FileAccPropList::DEFAULT;
+  FileAccPropList fapl = FileAccPropList::DEFAULT;
+  H5Pset_libver_bounds(fapl.getId(), H5F_LIBVER_LATEST, H5F_LIBVER_LATEST);
```
oruebel, line 53: What is the minimum version of HDF5 that we require for this?

stephprince: I believe HDF5 1.10

src/nwb/NWBFile.hpp
```diff
@@ -46,12 +46,12 @@ class NWBFile
    * @brief Initializes the NWB file by opening and setting up the file
    * structure.
    */
-  void initialize();
+  Status initialize();
```
oruebel, line 49: Please add a comment to the docs of all functions that are not accessible when the file is in recording mode, i.e., `initialize`, `finalize` etc.. Maybe something along the lines of: "If the file is in recording mode (i.e., when isRecording is true), then this function will fail. To modify the structure of the NWB file (i.e., add datasets, groups, etc.) the recording must be paused or stopped)"

src/hdf5/HDF5IO.cpp
stephprince: That sounds good, linked a low priority issue on this.

oruebel:
I think `SWMR` mode requires some user documentation, both because users are likely not familiar with this and because it provides AqNWB a set of unique features that have been called by the community as problems (in particular avoiding corruption of HDF5 files and being able to read while recording. 

I filed a separate issue for this https://github.com/NeurodataWithoutBorders/aqnwb/issues/48 to  make sure we document: a) what it is, b) what it means for the user (i.e., we can read during recording and avoid file corruption), c) what the use pattern of the API is.

src/nwb/NWBFile.cpp
```diff
@@ -30,19 +30,23 @@ NWBFile::NWBFile(const std::string& idText, std::shared_ptr<BaseIO> io)
...
-    createFileStructure();
+    return createFileStructure();
   }
 }
 
-void NWBFile::finalize()
+Status NWBFile::finalize()
 {
-  io->close();
+  return io->close();
```
oruebel, line 49: Should `finalize` also need to call `stopRecording`?

oruebel, line 49: Should `finalize` also clear the `recordingContainers`?

src/nwb/NWBFile.hpp
```diff
@@ -121,6 +126,12 @@ class NWBFile
   void cacheSpecifications(const std::string& specPath,
                            const std::string& versionNumber);
 
+  /**
+   * @brief Whether the file is in recording mode (can write / append to
+   * existing datasets but cannot create new groups or datasets)
+   */
```
oruebel, line 132: Note, this is actually a HDF5-specific limitation. In Zarr this is not an issue because all objects are stored in separate directories. I think instead of using `isRecording` in the `NWBFIle` to determine if we can create new objects it would be best to ask the io backend if the operation is allowed or not, i.e., we could have a function in `BaseIO` (overwritten in `HDF5IO`) that says something like `canUpdateFileStructure()`. When SWMR mode is enable or we are in 'read' mode that function would return `False` and `True` otherwise. In this way we can avoid having HDF5-specific limitations trickle into the NWB API classes.

oruebel: I think this also means that the `isRecording` property should be on `BaseIO` and set it in `BaseIO.startRecording` and `HDF5IO.startRecording`

src/BaseIO.hpp
```diff
@@ -238,6 +238,12 @@ class BaseIO
   virtual Status createReferenceDataSet(
       const std::string& path, const std::vector<std::string>& references) = 0;
 
+  /**
+   * @brief Starts the recording process.
+   * @return The status of the operation.
+   */
+  virtual Status startRecording() = 0;
```
oruebel, line 245: We should also add a `stopRecording` method on `BaseIO` and `HDF5IO`. Otherwise we can start a recording in the IO but not stop it. This is also goes along with the comment about moving the `isRecording` flag to the IO.

tests/testHDF5IO.cpp
```diff
@@ -274,4 +264,35 @@ TEST_CASE("writeAttributes", "[hdf5io]")
...
+  std::vector<SizeType> dataShape = {numSamples};
+  std::vector<SizeType> positionOffset = {0};
+  dataset->writeDataBlock(
+      dataShape, positionOffset, BaseDataType::I32, &testData[0]);
+  H5Dflush(static_cast<HDF5::HDF5RecordingData*>(dataset.get())
+               ->getDataSet()
+               ->getId());
+
+  // close file
+  hdf5io->close();
```
oruebel, line 297: It would be good to assert here that the data has been written successfully after the `hdf5io->close()`.

tests/testHDF5IO.cpp
```diff
@@ -274,4 +264,35 @@ TEST_CASE("writeAttributes", "[hdf5io]")
...
+  REQUIRE(status == Status::Success);
+
+  // write data block and flush
+  std::vector<SizeType> dataShape = {numSamples};
+  std::vector<SizeType> positionOffset = {0};
+  dataset->writeDataBlock(
+      dataShape, positionOffset, BaseDataType::I32, &testData[0]);
+  H5Dflush(static_cast<HDF5::HDF5RecordingData*>(dataset.get())
+               ->getDataSet()
+               ->getId());
```
oruebel, line 294: Why is the `H5Dflush` needed here? Shouldn't `hdf5io->close()` take care of flushing the data to disk?

tests/testHDF5IO.cpp
```diff
@@ -274,4 +264,35 @@ TEST_CASE("writeAttributes", "[hdf5io]")
...
+
+  // close file
```
oruebel, line 296: To check that the file can be opened again, we could potentially check after the `H5Dflush` and before the `hdf5io->close` whether the data can be read by opening the file again with HDF5 in read mode

oruebel (changes requested):
Overall this looks good. I think the main code item is to move IO specific properties to the IO backend to avoid backend-specific features in the NWB classes, i.e.,
      - track the `isRecording` status in the IO backend
      - have a way for the IO backend to say if objects can be created or not

src/nwb/NWBFile.cpp
oruebel: Resolving this here since it is part of #46

stephprince: I think that makes sense to call `stopRecording` in `finalize` in case the user has not already called it. Updated this accordingly.

src/nwb/NWBFile.hpp
stephprince: > I think this also means that the isRecording property should be on BaseIO and set it in BaseIO.startRecording and HDF5IO.startRecording. 

I agree that makes sense, I've updated accordingly. 

> Note, this is actually a HDF5-specific limitation. In Zarr this is not an issue because all objects are stored in separate directories. I think instead of using isRecording in the NWBFIle to determine if we can create new objects it would be best to ask the io backend if the operation is allowed or not, i.e., we could have a function in BaseIO (overwritten in HDF5IO) that says something like canUpdateFileStructure(). When SWMR mode is enable or we are in 'read' mode that function would return False and True otherwise. In this way we can avoid having HDF5-specific limitations trickle into the NWB API classes.

I see what you are saying about having the HDF5-specific limitations trickle into the main classes. Currently the `isRecording` flag is really only being used to detect whether we can create new objects so I replaced in entirely with `canModifyObjects`. 

I think if we establish a general usage pattern of 1. setup file structure, 2. create all datasets 3. record to datasets 4. stop recording and perform final edits / cleanup, it could be simpler to keep the name as `isRecording`? `createFileStructure` and `create{X}Series` are the main functions that currently need to consider this property and at least `createFileStructure` should probably not be called mid recording? But if `canModifyObjects` or some naming like that is clearer as to the intention, and if adding new objects during the recording process is something that Zarr / non-SWMR mode HDF5 files would want to do, then that works.

tests/testHDF5IO.cpp
stephprince: Yes as you mentioned below, I was using H5Dflush so that I could manually check that the data could be read by a different process before closing the file.

I've been trying to setup the tests to check that reading the data would fail if the file was not in SWMR mode. However, this seems to be more complicated than I expected, as it turns out that reading when the file is not in SWMR mode will work fine as long as it's in the same process, even if I set up multiple threads (see [related discussion here](https://forum.hdfgroup.org/t/swmr-r-w-in-two-processes-good-in-same-process-bad/10464/6)).

I was trying to setup multiple processes with `fork`, but it turns out Catch2 has some [known limitations](https://github.com/catchorg/Catch2/blob/devel/docs/limitations.md) related to this. You suggested a third option of calling a command line script to read from the file above, so I can try doing that instead.

src/nwb/NWBFile.hpp
stephprince: Added this text: "Note, this function will fail if the file is in a mode where new objects cannot be added"

src/nwb/NWBFile.hpp
oruebel: > I see what you are saying about having the HDF5-specific limitations trickle into the main classes. Currently the `isRecording` flag is really only being used to detect whether we can create new objects so I replaced in entirely with `canModifyObjects`.

That looks good to me. 

> I think if we establish a general usage pattern of 1. setup file structure, 2. create all datasets 3. record to datasets 4. stop recording and perform final edits / cleanup, it could be simpler to keep the name as `isRecording`?

For now, I'm leaning towards keeping it as `canModifyObjects`, mainly because this is really just a restriction of `SWMR` mode and will likely not be needed by backends like Zarr. I agree that we want to promote this pattern as a best practice, but I'm not sure right now if want to make it a hard restriction yet for all backends.

tests/testHDF5IO.cpp
oruebel: > You suggested a third option of calling a command line script to read from the file above, so I can try doing that instead

Since this is for improving testing, I think it would be find to address this in a separate issue, i.e., we shouldn't block merging this PR because of this particular issue since we know it is working.

src/hdf5/HDF5IO.cpp
```diff
@@ -377,6 +379,21 @@ Status HDF5IO::createStringDataSet(const std::string& path,
...
+  if (H5Fstart_swmr_write(this->file->getId()) < 0) {
+    return Status::Failure;
+  }
+  return Status::Success;
+}
+
+Status HDF5IO::stopRecording()
+{
+  modifyObjects = true;
+  return Status::Success;
```
oruebel, line 394: If I understand it correctly, this doesn't actually stop the `SWMR` mode, i.e., even though `modifyObjects` is set to `true` here, modifying objects would still fail. I don't think there is a `H5Fstop_swmr_write` function, so I think what we need to do here is:

- flush all data to disk
- close the HDF5 file
- open the HDF5 file again in `r+` mode

src/BaseIO.hpp
```diff
@@ -328,6 +348,13 @@ class BaseIO
    * @brief Whether the file is currently open.
    */
   bool opened;
+
+  /**
+   * @brief Whether the file is in a mode where objects can
+   * be added or deleted. Note, this does not apply to the modification
+   * of raw data on already existing objects.
+   */
+  bool modifyObjects = true;
```
oruebel, line 357: I don't think we need the `modifyObjects` as a variable here. Instead, I think we can just have `canModifyObjects` return `true` in `BaseIO` and then `HDF5IO` and other backends can just overwrite `canModifyObjects`. The main reason is, that this will allow backends to implement logic to check for `canModifyObjects` instead of forcing the backend to maintain the status of the boolean flag.

src/hdf5/HDF5IO.cpp
stephprince: Good point. Yes, I don't think there is a `H5Fstop_swmr_write` function, the recommended usage is to close and open again in `r+` mode.

We discussed that there is likely a distinction between pausing and stopping a recording (https://github.com/NeurodataWithoutBorders/aqnwb/issues/28), but I'm not sure yet what logic we would want to implement for each. I think it makes sense for now for `stopRecording` to close / reopen the file, and we can think about any distinctions for `pauseRecording` when I address that issue.

src/hdf5/HDF5IO.cpp
```diff
@@ -377,6 +379,21 @@ Status HDF5IO::createStringDataSet(const std::string& path,
   return Status::Success;
 }
 
+Status HDF5IO::startRecording()
+{
+  modifyObjects = false;
```
oruebel, line 384: Related to the above comment, I think in `HDF5IO`, instead of having to set the `modifyObjects` flag manually, we could implement `canModifyObjects` to check if `SWMR` mode is enabled. In this way we can avoid potential inconsistencies between the mode of the file and the status of `modifyObjects` flag. I.e., I think `canModifyObjects` would looks as follows:


```
bool HDF5IO::canModifyObjects()
{
   // Get the file access intent
    unsigned int intent;
    herr_t status = H5Fget_intent(file_id, &intent);
    if (status < 0) {
        return false;  // We could not access the file so modifying objects is not going to work
    }

    // Check if SWMR mode is enabled
    if (intent & H5F_ACC_SWMR_READ) {
        return False;  // File is in SWMR mode
    } else {
        return true;  // File is not in SWMR mode
    }
}
```

src/nwb/NWBFile.hpp
```diff
@@ -46,26 +46,33 @@ class NWBFile
...
    * @brief Finalizes the NWB file by closing it.
    */
-  void finalize();
+  Status finalize();
 
   /**
    * @brief Create ElectricalSeries objects to record data into.
    * Created objects are stored in recordingContainers.
+   * Note, this function will fail if the file is in a mode where
+   * new objects cannot be added.
```
oruebel, line 60:
```suggestion
   * Note, this function will fail if the file is in a mode where
   * new objects cannot be added which can be checked via
   * nwbfile.io->canModifyObjects()
```

oruebel: I would also add a similar comment to the other `create` methods where this applies

src/nwb/NWBFile.hpp
```diff
@@ -95,6 +102,8 @@ class NWBFile
 protected:
   /**
    * @brief Creates the default file structure.
+   * Note, this function will fail if the file is in a mode where
+   * new objects cannot be added.
```
oruebel, line 106:
```suggestion
   * Note, this function will fail if the file is in a mode where
   * new objects cannot be added. which can be checked via
   * nwbfile.io->canModifyObjects()
```

src/nwb/NWBRecording.cpp
```diff
@@ -30,13 +30,15 @@ Status NWBRecording::openFile(const std::string& rootFolder,
                                            createIO(IOType, filename));
   nwbfile->initialize();
 
+  // create the datasets
+  nwbfile->createElectricalSeries(recordingArrays);
+
   // start the new recording
-  return nwbfile->createElectricalSeries(recordingArrays);
+  return nwbfile->startRecording();
```
oruebel, line 37: For now this is fine, but we should discuss in a separate issue how we want to structure the `openFile`. Currently this assumes that a user wants to setup an exephys recording with just ElectricalSeries. I think as we add more types, I think it would be easier to have the user call these manually, i.e,.a user would call:

1. `NWBRecording::openFile` (which would just initalize the file)
2. call `nwbrecording->nwbfile->createElectricalSeries` (and any other create methods needed)
3. call `nwbrecroding->startRecording`
```

File: tests/testHDF5IO.cpp [hunk 16]
```diff
@@ -274,4 +264,24 @@ TEST_CASE("writeAttributes", "[hdf5io]")
...
+  std::vector<int> testData = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+  std::string dataPath = "/data";
+  SizeType numSamples = testData.size();
+  std::unique_ptr<BaseRecordingData> dataset = hdf5io->createArrayDataSet(
+      BaseDataType::I32, SizeArray {0}, SizeArray {1}, dataPath);
+
+  // turn on swmr mode
+  Status status = hdf5io->startRecording();
+  REQUIRE(status == Status::Success);
+  hdf5io->close();
```
oruebel, line 286:
```suggestion
  status = hdf5io->stopRecording();
  REQUIRE(hdf5io->canModifyObjects() == true);
  hdf5io->close();
```

tests/testNWBFile.cpp
```diff
@@ -29,9 +29,13 @@ TEST_CASE("createElectricalSeries", "[nwb]")
...
 
-  // start recording
+  // create Electrical Series
   std::vector<Types::ChannelVector> mockArrays = getMockChannelArrays(1, 2);
-  Status result = nwbfile.createElectricalSeries(mockArrays, BaseDataType::F32);
+  Status resultCreate =
+      nwbfile.createElectricalSeries(mockArrays, BaseDataType::F32);
+
+  // start recording
+  Status resultStart = nwbfile.startRecording();
```
oruebel, line 38:
```suggestion
  Status resultStart = nwbfile.startRecording();
  REQUIRE(resultStart == Status::Success);
```

tests/testNWBFile.cpp
```diff
@@ -48,5 +52,32 @@ TEST_CASE("createElectricalSeries", "[nwb]")
 
   nwbfile.finalize();
 
-  REQUIRE(result == Status::Success);
+  REQUIRE(resultCreate == Status::Success);
```
oruebel, line 55: Why is `resultCreate` checked here and not above where it is being set?

tests/testNWBFile.cpp
```diff
@@ -48,5 +52,32 @@ TEST_CASE("createElectricalSeries", "[nwb]")
 
   nwbfile.finalize();
```
oruebel, line 53:
```suggestion
  Status finalized nwbfile.finalize();
  REQUIRE(finalized == Status::Success);
```

It would also be useful to have some checks that the data that is being written here is correct, or that being checked elsewhere?

oruebel (changes requested):
I added few more comments, but overall this looks good.

src/hdf5/HDF5IO.cpp
oruebel: > We discussed that there is likely a distinction between pausing and stopping a recording (#28), but I'm not sure yet what logic we would want to implement for each.

I think the current version of the code already implements this in `NWBRecording` in that:

* `NWBFile.stopRecording` means to stop the recording process, but leave the file open (so that a user can continue later on)
* `NWBFile.finalize` means to stop the recording and close the file (i.e., a true stop)

In that sense, we could rename `stopRecording` to `pauseRecording` in both `NWBFile` and `baseIO` if that is more intuitive.

src/nwb/NWBFile.cpp
```diff
@@ -131,7 +140,15 @@ Status NWBFile::createElectricalSeries(
...
-void NWBFile::stopRecording() {}
+Status NWBFile::startRecording()
+{
+  return io->startRecording();
+}
+
+void NWBFile::stopRecording()
+{
+  io->stopRecording();
+}
```
oruebel, line 151: Can you clarify why `startRecording`, `stopRecording` are part of `NWBFile` and not `NWBRecording`? Since `NWBRecording` manages the recording process, those functions seem to logically belong there (or maybe they are needed on both `NWBFile` and `NWBRecording`, depending on if `NWBFile` has to do anything special here?)

tests/testHDF5IO.cpp
stephprince: I added a test that checks the dataset size increases after writing in swmr mode.

tests/testHDF5IO.cpp
stephprince: I added a test that checks the file cannot be read before starting swmr mode and can be read after `startRecording`

src/hdf5/HDF5IO.cpp
stephprince: > think the current version of the code already implements this in NWBRecording

That's true. I changed the name to `pauseRecording`. 

I also added a test to check that the `recordingContainers` time series are still accessible after pausing the recording - I think that would be the other main difference between stopping and pausing?

tests/testNWBFile.cpp
stephprince: I think that was a mistake when I added additional code, I moved it up to where it is being set

tests/testNWBFile.cpp
stephprince: I have some checks for content as part of the `testNWBRecording.cpp` and some general checks for data block write in `testHDF5IO.cpp`. However it's probably a good idea to add additional checks for each data type.

tests/testHDF5IO.cpp
stephprince: I think I got it working!

src/nwb/NWBFile.cpp
stephprince: I think currently the main reason it's part of `NWBFile` is because the `io` object is a private member variable of `NWBFile`. 

As you mentioned above, currently `NWBRecording::openFile` assumes a specific recording process and automatically sets that up, calling `NWBFile::startRecording` at the end. However I agree if we want `NWBRecording` to manage the recording process and want to allow more customizability for different datatypes, it makes sense to at least have `startRecording` on both.

src/hdf5/HDF5IO.cpp
stephprince: That makes sense to avoid inconsistencies. I added a `canModifyObjects` function with a note for derived classes to override the `BaseIO` implementation if needed

src/hdf5/HDF5IO.cpp
stephprince: On second thought - I'm not sure they should still be accessible, and I think this might be preventing the file from actually closing properly? I'm running into issues reopening the file during the `pauseRecording` function without `H5F_ACC_SWMR_WRITE` intent, and I think this may be the issue.

src/nwb/NWBFile.cpp
oruebel: > I think currently the main reason it's part of `NWBFile` is because the `io` object is a private member variable of `NWBFile`.

Let's discuss the design for `NWBRecording`. I'm wondering whether the `io` should be owned by `NWBRecording`, but that'll depend on where we decide to draw the line in terms of what the responsibility of `NWBFile` and `NWBRecording` are. But let's deal with that in a separate PR.

src/BaseIO.hpp
```diff
@@ -238,6 +238,27 @@ class BaseIO
...
       const std::string& path, const std::vector<std::string>& references) = 0;
 
+  /**
+   * @brief Starts the recording process.
+   * @return The status of the operation.
+   */
+  virtual Status startRecording() = 0;
+
+  /**
+   * @brief Stops the recording process.
```
oruebel, line 248:
```suggestion
   * @brief Pauses the recording process.
   * The file should remain open to support modifications and 
   * restart of the recording by calling startRecording again.
```

tests/testNWBFile.cpp
stephprince: I forgot that I had added checks for `ElectricalSeries` in `testEcephys.cpp`. I think that should cover these cases for now.

src/hdf5/HDF5IO.cpp
oruebel: > I also added a test to check that the `recordingContainers` time series are still accessible after pausing the recording - I think that would be the other main difference between stopping and pausing?

Yeah, I think that's a problem. When closing and reopening the file, I believe the recordingContainers are all going to be invalid, because the HDF5 objects they point to depend on the file object that was opened before. The only way to really prevent this from happending would be if the recordingContainers (or any NWB object) would not store the object but request them from the IO object each time. Let's discuss tomorrow how we want to address this. Maybe we should just leave it as a true stop of the recording for now, i.e., no restart. Once we add read, we could then do a restart by opening the file again.

oruebel: Other than this last detail, the PR looks good.

stephprince:
We discussed offline that we wanted to implement the following changes, these should also fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/47 by making swmr mode optional.
- Have HDF5IO take an optional `disableSWMRMode` argument (default `false`)
- Have `startRecording` and `stopRecording` detect whether SWMR mode is enabled
- Change `pauseRecording` back to `stopRecording`. We will likely not have full pauseRecording capabilities in swmr mode (i.e. the ability to pause, add additional objects, and restart the recording) until we implement the read side of the API.
- Check in other io functions whether the file is open before performing any tasks

I have left out any changes to `NWBRecording` inputs for now, since we are still deciding the structure of `NWBRecording` vs. `NWBFile`.

src/hdf5/HDF5IO.cpp
```diff
@@ -377,6 +386,49 @@ Status HDF5IO::createStringDataSet(const std::string& path,
...
+  }
+  return Status::Success;
+}
+
+Status HDF5IO::stopRecording()
+{
+  // if SWMR mode is disabled, stopping the recording will leave the file open
+  if (!disableSWMRMode) {
+    close();
+  }
```
oruebel, line 406:
```suggestion
  if (!disableSWMRMode) {
    close();  // SWMR mode cannot be disabled so close the file
  }
  else
  {
    H5Fflush(this->file->getId(), H5F_SCOPE_GLOBAL); // flush all data to disk
  }
```

src/hdf5/HDF5IO.cpp
```diff
@@ -377,6 +386,49 @@ Status HDF5IO::createStringDataSet(const std::string& path,
...
+
+  // Check if SWMR mode is enabled
+  if (intent & (H5F_ACC_SWMR_READ | H5F_ACC_SWMR_WRITE)) {
+    return false;  // File is in SWMR mode
+  } else {
+    return true;  // File is not in SWMR mode
+  }
+  return true;
+}
+
```
oruebel, line 431: I think the logic in this function could be simplified to reduce the number of return statements. I think this should work:

```suggestion
{
  // check if we are in swmr mode
  bool smwr_mode = False
  unsigned int intent;
  herr_t status = H5Fget_intent(this->file->getId(), &intent);
  bool status_ok = (status >= 0)
  if(status_ok)
  {
      swmr_mode = (intent & (H5F_ACC_SWMR_READ | H5F_ACC_SWMR_WRITE))
  }
  // if the file is opened and we are not in swmr mode then we can modify objects
  return opended && status_ok && !swmr_mode
}
```

What do you think?

src/hdf5/HDF5IO.cpp
```diff
@@ -48,15 +50,16 @@ Status HDF5IO::open(bool newfile)
...
+  H5Pset_libver_bounds(fapl.getId(), H5F_LIBVER_LATEST, H5F_LIBVER_LATEST);
 
   if (newfile)
     accFlags = H5F_ACC_TRUNC;
   else
     accFlags = H5F_ACC_RDWR;
 
   file = std::make_unique<H5::H5File>(
-      getFileName(), accFlags, FileCreatPropList::DEFAULT, props);
+      getFileName(), accFlags, FileCreatPropList::DEFAULT, fapl);
```
oruebel, line 62: In line 70 in `HDF5IO::close()`, it looks like the function doesn't actually close the HDF5 file. I think it should call:

```
    if(this->file != nullptr && opened)
    {
       H5Fclose(this->file->getId());
    }
``

File: src/hdf5/HDF5IO.hpp [hunk 25]
```diff
@@ -232,6 +252,7 @@ class HDF5IO : public BaseIO
 
 private:
   std::unique_ptr<H5::H5File> file;
+  bool disableSWMRMode;
```
oruebel, line 255:
```suggestion
  bool disableSWMRMode;   // when set do not use SWMR mode when opening the HDF5 file
```

src/nwb/NWBFile.cpp
```diff
@@ -30,24 +30,28 @@ NWBFile::NWBFile(const std::string& idText, std::shared_ptr<BaseIO> io)
 
 NWBFile::~NWBFile() {}
 
-void NWBFile::initialize()
+Status NWBFile::initialize()
```
oruebel, line 33: I think it may be useful to add a `bool` flag here for `overwriteIfExists`, but than can be a seperate low-priority issue

tests/testHDF5IO.cpp
```diff
@@ -274,4 +269,120 @@ TEST_CASE("writeAttributes", "[hdf5io]")
...
+      for (size_t i = 0; i < testData.size(); ++i) {
+        testData[i] += (testData.size());
+      }
+    }
+
+    // stop recording, check that file is still open and recording can be
+    // restarted
+    status = hdf5io->stopRecording();
+    REQUIRE(hdf5io->isOpen() == true);
+    REQUIRE(hdf5io->startRecording() == Status::Success);
```
oruebel, line 386: In addition to the checking that the file is open, I think it would be useful to check here that you can: 1) create a new ElectricalSeries (or some other group), 2) write new data to the file again by starting the recording again and writing data blocks.

oruebel (changes requested):
A couple of minor additional comments but otherwise this looks good

src/hdf5/HDF5IO.cpp
stephprince: I think we still need the early return for `!opened` because `this->file` will be a `nullptr`, but otherwise that works!

src/hdf5/HDF5IO.cpp
stephprince: updated using the `file->close()` function, I think they are equivalent.

stephprince:
Finished updating based on the latest suggestions!

oruebel (approved):
Looks good. Thanks for your patients!

PR #58, opened 2024-08-02
Show figure in README also in Doxygen docs

### LINKED ISSUE #57: Figure in README.md missing in Doxygen output
The figure in the README.md file does not show up in the Doxygen HTML. The issue seems to be that the image file is not correctly being copied by Doxygen. For details see this issue https://github.com/doxygen/doxygen/issues/6783

To work around this issue we could scale the image we want to show beforehand so that we can use the `Project Overview`, which should work in both GitHub and Doxygen. Having to scale the figures is not ideal, but it seems like the simplest solution. Otherwise (it seems) we'll need to manually copy the images to Doxygen ourselves or place all images in the same folder as the markdown file, which would clutter the root folder of the repo.

oruebel:
Fix #57

The figure in the README.md file does not show up in the Doxygen HTML. The issue seems to be that the image file is not correctly being copied by Doxygen. For details see this issue https://github.com/doxygen/doxygen/issues/6783

To work around this issue this PR adds a scaled version of the figures so that we can use the markdown syntax 
`Project Overview` instead of HTML syntax (but the markdown syntax does not allow for setting the figure size, so we need to scale the image).


stephprince (approved):
Sounds good, looks good to me!

PR #56, opened 2024-08-02
Update docs to fix Doxygen build warnings

oruebel:
Fix #50

Update docstrings to fix warnings in Doxygen build

PR #59, opened 2024-08-04
Add action to deploy docs via GitHub pages

### LINKED ISSUE #24: Set up documentation page
Set up readthedocs page using doxygen documentation.

oruebel:
Fix #24 

Add GitHub action to automatically update and deploy the Doxygen documentation via GitHub pages on every push to the main branch

TODO:

- [x] Fix building the docs
- [ ] Add separate action to test docs build on PRs (without deploying the docs and with doxygen warnings elevated to errors) See the discussion here https://github.com/NeurodataWithoutBorders/aqnwb/pull/56#issuecomment-2269387480. The Doxygen settings to elevate warnings are here https://www.doxygen.nl/manual/config.html#cfg_warn_as_error. I think we want to set `WARN_AS_ERROR = FAIL_ON_WARNINGS_PRINT ` so that Doxygen runs to the end, prints warnings as a single block, but then exits with a an error code. In this way we can see all the warnings (instead of failing after the first one) and all warnings are listed together to make them easier to review.


oruebel:
@stephprince the main structure for the CI to deploy the docs on GitHub pages is here, but I ran into problems building the docs via cmake in the CI https://github.com/NeurodataWithoutBorders/aqnwb/actions/runs/10232814207/job/28310386149?pr=59

The problem is here:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/32e8b7d82c0462c527792061fde1f3fca9095979/.github/workflows/doxygen-gh-pages.yml#L36

and seems related to the `ci-ubuntu` preset. I tried to follow the instructions in the Readme but the `dev` preset does not exist and I'm not sure what the right preset is to use for building the docs. Otherwise, the action should be fine.

.github/workflows/doxygen-gh-pages.yml
```diff
@@ -0,0 +1,48 @@
+name: Doxygen GitHub Pages Deploy Action
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
oruebel, line 7:
```suggestion
```

Remove the `on: pull_request` option once the PR is ready. This option is only used for debugging so that the Action runs on this PR.

.github/workflows/doxygen-gh-pages.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Create .nojekyll to ensure pages with underscores work on gh pages
+        run: touch build/docs/html/.nojekyll
+        shell: bash
+
+      - name: Deploy to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: test_gh_pages
```
oruebel, line 47:
```suggestion
          branch: gh_pages
```

Change the target branch to `gh_pages` once ready. It is set to a throwaway branch here for testing only.

stephprince:
> the main structure for the CI to deploy the docs on GitHub pages is here, but I ran into problems building the docs via cmake in the CI 

I'm not sure if the separate `docs-ci.cmake` file is necessary for our use, but it was one of the files already setup by the cmake project initializer. I updated the workflow to use that implementation.


> I tried to follow the instructions in the Readme but the dev preset does not exist and I'm not sure what the right preset is to use for building the docs. Otherwise, the action should be fine.

The dev preset is a user-specific preset that inherits from `"dev-mode", "ci-<os>"` from `CmakePresets.json` (depending on your local os). I can add instructions for setting that up (see [here](https://github.com/friendlyanon/cmake-init-shared-static/blob/c4e38d25f8d384e19c98bc17dbb0b84982444788/HACKING.md?plain=1#L32) for the example I followed).

oruebel:
> The dev preset is a user-specific preset that inherits from `"dev-mode", "ci-<os>"` from `CmakePresets.json` (depending on your local os). I can add instructions for setting that up (see [here](https://github.com/friendlyanon/cmake-init-shared-static/blob/c4e38d25f8d384e19c98bc17dbb0b84982444788/HACKING.md?plain=1#L32) for the example I followed).

I would at least link to those instructions in the README or just add the `dev` preset as a standard preset

oruebel:
> I had updated the action to use the `docs-ci.cmake` version and it was working, but seems to be failing again with the latest merges, I can look into it.

Looking at the CI, I think the path where the docs are being built has changed so we just need to update this line

https://github.com/NeurodataWithoutBorders/aqnwb/blob/e910f239a5de02f36c41c51095a4b088aad00b7f/.github/workflows/doxygen-gh-pages.yml#L40

to create the `.nojekyll` in the correct location

oruebel:
> I added instructions for the dev preset to the README. Since it will change for each user, I will leave it off the standard presets list for now.

Thanks! That looks good.

oruebel:
@stephprince do we actually need to compile AqNWB to build the docs? If not, then I think the Action could be simplified a bit by removing install of libraries not needed for the docs. But that's not critical.

https://github.com/NeurodataWithoutBorders/aqnwb/blob/e910f239a5de02f36c41c51095a4b088aad00b7f/.github/workflows/doxygen-gh-pages.yml#L17-L33

oruebel:
Oh, and the path in the deploy part will need to be updated as well:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/e910f239a5de02f36c41c51095a4b088aad00b7f/.github/workflows/doxygen-gh-pages.yml#L48

PR #62, opened 2024-08-05
Docs doxygen style update

oruebel:
This PR uses the [doxygen-awesome](https://github.com/jothepro/doxygen-awesome-css) style to customize the look of the docs to give them a bit more modern style. Main changes include:

- [X] Add `doxygen-awesome-sidebar-only.css` and `doxygen-awesome.css` files 
- [X] Update  `Doxyfile.in` to set the HTML style for a sidebar-only style layout
- [X] Update  `Doxyfile.in` to use GraphViz to generate interactive vector graphics for the UML diagrams (instead of the default, low-res PNGs)
- [X] Add Doxygen default `header.html` so that we can customize the header 
- [X] Add draft logo to the header

Screen Shot 2024-08-04 at 6 20 16 PM


stephprince (approved):
Looks great! I like the more modern style.

PR #64, opened 2024-08-06
fix electricalSeries default inputs

### LINKED ISSUE #53: Why is `electrodesTablePath` needed?
https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/nwb/ecephys/ElectricalSeries.cpp#L12

is required when creating an `ElectricalSeries`. It is unclear to me why is this is needed? The NWB schema specifies a single `ElectrodesTable` that is in a fixed location in the file. As such, the `electrodesTablePath` should be the same for all NWB files.

### LINKED ISSUE #54: `ElectricalSeries.units` should not be setable
https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/nwb/ecephys/ElectricalSeries.hpp#L30

provides a default value of `volts` but allows the user to change the untis. However, in the `NWBSchema` the value for `units` are fixed to `volts`, and as such no other value is allowed. We should, hence, not allow the user to set this value but fix it to `volts` in the code.

### LINKED ISSUE #55: Invalid default values for `ElectricalSeries.dsetSize` and `ElectricalSeries.chunkSize`
https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/nwb/ecephys/ElectricalSeries.hpp#L32-L33

specify vectors with a single element as the default parameter. However, the `ElectricalSeries` data must always be 2D and the second dimension must have the same length as the number of electrods/channels. As such, the default values are not valid, potentially resulting in the creation of an invalid file. I think we can simply remove the default values and require that they must be specified.

stephprince:
Fix several issues related to the inputs to `ElectricalSeries` in preparation for new data types that will inherit from ElectricalSeries.
* fix #53 
* fix #54 
* fix #55

Regarding #53, I didn't realize the path was specified in the schema and there is only a single `ElectrodesTable` - that makes things easier!

### DIFF
--- a/src/nwb/ecephys/ElectricalSeries.hpp
+++ b/src/nwb/ecephys/ElectricalSeries.hpp
@@ -75,11 +75,6 @@ public:
   Types::ChannelVector channelVector;
 
   /**
-   * @brief Path to the electrodes table this time series references
-   */
-  std::string electrodesTablePath = "/general/extracellular_ephys/electrodes/";
-
-  /**
    * @brief Pointer to channel-specific conversion factor dataset.
    */
   std::unique_ptr<BaseRecordingData> channelConversion;
--- a/src/nwb/ecephys/ElectricalSeries.hpp
+++ b/src/nwb/ecephys/ElectricalSeries.hpp
@@ -20,8 +20,6 @@ public:
    * @param io A shared pointer to the IO object.
    * @param dataType The data type to use for storing the recorded voltage
    * @param channelVector The electrodes to use for recording
-   * @param electrodesTablePath Path to the electrodes table
-   * @param unit Unit for the electrical signal. Must be "volts".
    * @param description The description of the TimeSeries.
    * @param dsetSize Initial size of the main dataset. This must be a vector
    *                 with two elements. The first element specifies the length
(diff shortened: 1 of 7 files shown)


src/nwb/ecephys/ElectricalSeries.hpp
```diff
@@ -81,7 +77,7 @@ class ElectricalSeries : public TimeSeries
   /**
    * @brief Path to the electrodes table this time series references
    */
-  std::string electrodesTablePath;
+  std::string electrodesTablePath = "/general/extracellular_ephys/electrodes/";
```
oruebel, line 80:
```suggestion
  inline static const std::string electrodesTablePath = "/general/extracellular_ephys/electrodes/";
```

oruebel: Actually, I would make `electrodeTablePath` a `static const` member of `ElectrodeTable` such that we can just grab from the `ElectrodeTable` class, rather than making it a property of `ElectricalSeries`. Alternatively, it could also be a `static const` member of `NWBFile` since it really defines where the `ElectrodesTable` goes.

tests/testFile.cpp
```diff
@@ -11,7 +11,7 @@ using namespace AQNWB;
 
 TEST_CASE("ElectrodeTable", "[ecephys]")
 {
-  std::string path = "/electrodes/";
+  std::string path = "/general/extracellular_ephys/electrodes/";
```
oruebel, line 14: If we make the `electrodeTablePath` a `const static` property then we can also avoid duplicating it here and the other places this path is defined, because it is then fixed at compile time.

oruebel (commented):
Other than the minor comment about moving `electrodeTablePath` and making it `const` this looks good.

src/nwb/ecephys/ElectricalSeries.hpp
stephprince: I see, yes that looks better and minimizes the path duplication. I implemented on `ElectrodeTable`

oruebel (approved):
Looks good to me

Issue #66, opened 2024-08-06
add workflow to test doxygen build


stephprince:
See discussion and commits in related PR here: https://github.com/NeurodataWithoutBorders/aqnwb/pull/59 

> Add separate action to test docs build on PRs (without deploying the docs and with doxygen warnings elevated to errors) See the discussion here https://github.com/NeurodataWithoutBorders/aqnwb/pull/56#issuecomment-2269387480. The Doxygen settings to elevate warnings are here https://www.doxygen.nl/manual/config.html#cfg_warn_as_error. I think we want to set WARN_AS_ERROR = FAIL_ON_WARNINGS_PRINT  so that Doxygen runs to the end, prints warnings as a single block, but then exits with a an error code. In this way we can see all the warnings (instead of failing after the first one) and all warnings are listed together to make them easier to review.

stephprince:
there are currently warnings likely related to the doxygen version - this should also be updated

PR #67, opened 2024-08-06
Add doxygen build test workflow

stephprince:
Fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/66.


stephprince:
Think this should work now - I removed the `M_SHOW_UNDOCUMENTED` tag because it is unsupported. The `EXTRACT_ALL` configuration should include all objects whether they are undocumented or not.

oruebel (approved):

PR #80, opened 2024-08-27
Add user docs and workflow

### LINKED ISSUE #48: Add user docs
Add documentation with information about the usage of the API. A few topics to cover here are:

Typical usage pattern of the API and order in which functions (e.g., in NWBFile should be called) to do a recording. E.g,.:

- setup the file structure
- record to the NWB file by adding date but not modifying the structure of the file
      - Also discuss which functions are  unaccessible during recording with HDF5 in SWMR mode. 
- pause/stop the recording to do (final) edits to the structure if necessary
- finalize the recording (and clean up objects)

Highlight features of the API
- Highlight that due to the use of SWMR mode:
     -  One can read from the file while it is being recorded to
     -  Ensures that the file does not get corrupted even if there are errors during the recording process (but we may lose any data that has not been flushed to disk yet)

oruebel:
Fix #48

Add description of the typical user workflow


oruebel:
Closing. This is being fixed in #84

Issue #38, opened 2024-07-27
Add tags to objects in recordingContainers


stephprince:
It may be useful to have convenience functions to group and find signals from recordingContainers using user-defined or automatically added tags. 

See further discussion here: https://github.com/NeurodataWithoutBorders/aqnwb/pull/33#discussion_r1693637985 and https://github.com/NeurodataWithoutBorders/aqnwb/pull/84#discussion_r1739301284

PR #84, opened 2024-08-30
Restructure classes and workflow

stephprince:
Based on our discussion earlier this week, we have decided to reorganize responsibilities among the main aqnwb classes.

These changes
- remove the `NWBRecording` class, the user will instead have control of the `NWBFile` object and the `openFile`/`closeFile` processes
- give the user control over I/O object creation to pass into the `NWBFile`
- move `RecordingContainers` to its own file, it will manage access / writing to datasets
- remove `startRecording` and `stopRecording` from the `NWBFile`, leave that responsibility to the I/O

I've also added documentation describing the workflow we discussed.


stephprince:
@oruebel I just noticed that you had started outlining the workflow in the other PR as well - we can try to merge them here or keep that in a separate PR?

oruebel:
> @oruebel I just noticed that you had started outlining the workflow in the other PR as well - we can try to merge them here or keep the documentation in a separate PR?

No worries about that. #80 is just a draft PR. I just wanted to keep notes about the workflow and I figured it was easiest to just put it in a draft PR so we have something to build off rather than bury the notes in a GoogleDoc. We can shepherd this PR through first. If #80 still has something useful to add after we are done with this PR, then I can work on that or otherwise we can just close #80.

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * For users wanting to integrate NWB with a particular data acquisition software, here
+ * we outline the steps for a single recording from file creation to saving.
+ *
+ * 1. Create the I/O object.
+ * 2. Create the `RecordingContainers` object. 
+ * 3. Create the `NWBFile` object.
+ * 4. Create datasets and add to RecordingContainers.
+ * 5. Start the recording.
+ * 6. Write data.
+ * 7. Stop the recording and close the NWBFile.
```
oruebel, line 17:
```suggestion
 * For users wanting to integrate NWB with a particular data acquisition software, here
 * we outline the steps for a single recording from file creation to saving.
 *
 * 1. Create the \ref AQNWB::HDF5::HDF5IO "HDF5IO" object used for writing data to the file on disk.
 * 2. Create the \ref AQNWB::NWB::RecordingContainers "RecordingContainers" object used for managing \ref AQNWB::NWB::Container "Container" objects for storing recordings.
 * 3. Create the \ref AQNWB::NWB::NWBFile "NWBFile" object used for managing and creating NWB file contents.
 * 4. Create the \ref AQNWB::NWB::Container "Container" objects (e.g., \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries") used for recording and add them to the \ref AQNWB::NWB::RecordingContainers "RecordingContainers".
 * 5. Start the recording.
 * 6. Write data.
 * 7. Stop the recording and close the \ref AQNWB::NWB::NWBFile "NWBFile".
```

Mainly changing names of classes to references and adding a bit of text to explain what each step/object is used for

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * 
+ * Below, we walk through these steps in more detail.
+ * 
+ * 
+ * \subsection create_io 1. Create the I/O object.
+ * 
+ * First, create an I/O object. We provide a convenience method, `createIO`
```
oruebel, line 24:
```suggestion
 * First, create an I/O object (e.g., \ref AQNWB::HDF5::HDF5IO "HDF5IO". We provide a convenience method, \ref AQNWB::createIO "createIO"
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * to create this object using one of our supported backends. For more fine-grained
+ * control of different backend parameters, you can create your own `std::shared_ptr` 
+ * using any of the derived BaseIO classes.
```
oruebel, line 27:
```suggestion
 * using any of the derived \ref AQNWB::BaseIO "BaseIO" classes.
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * 
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_io_snippet
+ *
+ *
+ * \subsection create_recording_container 2. Create the RecordingContainer object.
+ * 
+ * Next, create a RecordingContainer object to manage the different datasets
```
oruebel, line 34:
```suggestion
 * Next, create a \ref AQNWB::NWB::RecordingContainers "RecordingContainers"  object to manage the different \ref AQNWB::NWB::Container "Container" objects with the datasets
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * 
+ * Next, create a RecordingContainer object to manage the different datasets 
+ * that you would like to write data to.
+ * 
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_recording_containers_snippet
+ * 
+ *
+ * \subsection create_nwbfile 3. Create the NWBFile
+ * 
+ * Next, constructs the `NWBFile` object, using the I/O object as an input.
```
oruebel, line 42:
```suggestion
 * Next, constructs the \ref AQNWB::NWB::NWBFile "NWBFile"  object, using the I/O object as an input. 
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * 
+ * Next, constructs the `NWBFile` object, using the I/O object as an input. 
+ * Then, initialize the object to create the basic file structure of the NWBFile.
+ *
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_nwbfile_snippet
+ *
+ * 
+ * \subsection create_datasets 4. Create datasets and add to RecordingContainers.
+ *
+ * Next, create the different data types (e.g. `ElectricalSeries`, other `TimeSeries`)
```
oruebel, line 50:
```suggestion
 * Next, create the different data types (e.g. \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries" or other AQNWB::NWB::TimeSeries "TimeSeries")
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * that you would like to write data into. After creation, these objects are moved
+ * to the RecordingContainers object so that it can manage access and data writing
```
oruebel, line 52:
```suggestion
 * that you would like to write data into. After creation, these objects are added
 * to the  \ref AQNWB::NWB::RecordingContainers "RecordingContainers"  object so that it can manage access and data writing
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * during the recording process. 
+ * 
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_datasets_snippet
+ *
+ *
+ * \subsection start_recording 5. Start the recording.
+ *
+ * Then, start the recording process with a call to the I/O object. By default for
+ * HDF5 files, this will enable SWMR mode and no additional datasets or groups can be added
+ * to the file unless it is closed and reopened.
```
oruebel, line 62:
```suggestion
 * Then, start the recording process with a call to the ``startRecording`` function of the I/O object. 
 *
 *  \note 
 * When using \ref AQNWB::HDF5::HDF5IO "HDF5IO" for writing to HDF5, calling 
 * \ref AQNWB::HDF5::HDF5IO::startRecording "startRecording" will by default enable
 * \ref hdf5io_swmr "SWMR mode" to ensure file integrity and support concurrent read. 
 * As a result, no additional datasets or groups can be added to the file once a recording
 * has been started unless the file is is closed and reopened.
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_start_snippet
+ *
+ * 
+ * \subsection write_data 6. Write data.
+ *
+ * During the recording process, use the RecordingContainers as an interface
```
oruebel, line 69:
```suggestion
 * During the recording process, use the \ref AQNWB::NWB::RecordingContainers "RecordingContainers"  as an interface
```

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,83 @@
...
+ * to access the various datasets and write blocks of data to the file.
```
oruebel, line 70:
```suggestion
 * to access the various  \ref AQNWB::NWB::Container "Container" object and corresponding 
 * datasets and write blocks of data to the file. 
```

src/nwb/RecordingContainers.hpp
```diff
@@ -0,0 +1,74 @@
...
+   * @return The status of the write operation.
+   */
+  Status writeTimeseriesData(const SizeType& timeseriesInd,
+                             const Channel& channel,
+                             const std::vector<SizeType>& dataShape,
+                             const std::vector<SizeType>& positionOffset,
+                             const void* data,
+                             const void* timestamps);
+
+  std::vector<std::unique_ptr<TimeSeries>> containers;
```
oruebel, line 70:
```suggestion
  std::vector<std::unique_ptr<Container>> containers;
```

src/nwb/RecordingContainers.hpp
```diff
@@ -0,0 +1,74 @@
...
+   * @brief Adds a TimeSeries object to the container.
+   * @param data The TimeSeries object to add.
+   */
+  void addData(std::unique_ptr<TimeSeries> data);
+
+  /**
+   * @brief Gets the TimeSeries object from the recordingContainers
+   * @param timeseriesInd The index of the timeseries dataset within the group.
+   */
+  TimeSeries* getTimeSeries(const SizeType& timeseriesInd);
```
oruebel, line 48:
```suggestion
  Container* getContainer(const SizeType& containerInd);
```

tests/examples/testWorkflowExamples.cpp
```diff
@@ -0,0 +1,104 @@
...
+              dataBuffer.size(), channel.getBitVolts(), dataBuffer.data());
+
+          // [example_workflow_write_snippet]
+          recordingContainers->writeTimeseriesData(i,
+                                                   channel,
+                                                   dataShape,
+                                                   positionOffset,
+                                                   intBuffer.get(),
+                                                   timestampsBuffer.data());
+          // [example_workflow_write_snippet]
```
oruebel, line 89: Should we add an optional call to io->flush() so we can add a note in the docs that calling flush ensures that is moved to disk?

oruebel (changes requested):
Overall, this looks good to me. I add some comments on the workflow.dox, mainly to: 1) add a bit more detail about what the role of different classes are to clarify why they are needed and 2) change class names to references to make it easier for readers to find more information. The other high-level item is that I think `RecordingContainers` should accept all ``Container`` types not just ``TimeSeries``, we'll likely need non-TimeSeries for recording in the near future so I think it would be good to not restrict this.

src/nwb/RecordingContainers.hpp
oruebel: In addition to lookup by index, it would be nice to also have a way to look up object by name. This could be a separate issue. One option would be to: change  ``std::vector<std::unique_ptr<TimeSeries>> containers;`` to a Boost `multi_index_container` to allow for look-up both based on index of insertion and a name, where the name could either be custom , unique label defined by the user or the path of the object.

oruebel:
I closed #80 since you've covered everything that was in that PR here.

oruebel:
Fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/48

src/nwb/RecordingContainers.hpp
stephprince: > I think we probably want this to be Container. While we currently only use TimeSeries in recording, for event-based recordings (and I think also ROI's in ophys) this will be DynamicTable type containers so we should prepare for this. To safe-guard against issues, I'd add a check in writeTimeSeries to make sure the Container uses is actually a TimeSeries

I updated the class to apply to `Container` more generally. For checking whether the type is actually a `TimeSeries` or `ElectricalSeries`, I used dynamic casting since that should return a `nullptr` if the cast is invalid. Let me know if you had a different approach in mind here?

> In addition to lookup by index, it would be nice to also have a way to look up object by name.

I think this is related to the open issue we have to add tags to the recordingContainers so that we can look up by name/user-defined labels - https://github.com/NeurodataWithoutBorders/aqnwb/issues/38

oruebel: > I think this is related to the open issue we have to add tags to the recordingContainers so that we can look up by name/user-defined labels - #38

Agreed. I added a reference to #38 to this thread. 



> , I used dynamic casting since that should return a `nullptr` if the cast is invalid. Let me know if you had a different approach in mind here?

Since this involves dynamic lookup of objects, I think runtime checking with `dynamic_cast` is the right approach. For anything that we can already check at compile time, I would recommend using compile time asserts like ``static_assert(std::is_base_of<Base, T>::value, "T must be derived from Base")`` instead, but since the functions where this comes into play involve dynamic look-up of objects, I don't think compile time asserts are feasible (at least not without restricting functionality). Or in short, "looks good" 👍

src/nwb/RecordingContainers.hpp
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * @brief Destructor for RecordingContainer class.
+   */
+  ~RecordingContainers();
+
+  /**
+   * @brief Adds a Container object to the container.
+   * @param data The Container object to add.
+   */
+  void addData(std::unique_ptr<Container> data);
```
oruebel, line 42: I would rename `addData` to `addContainer` (just for consistency with the other names e.g. `getContainer` and `RecordingContainers`)

src/nwb/RecordingContainers.cpp
```diff
@@ -0,0 +1,65 @@
...
+using namespace AQNWB::NWB;
+// Recording Container
+
+RecordingContainers::RecordingContainers() {}
+
+RecordingContainers::~RecordingContainers() {}
+
+void RecordingContainers::addData(std::unique_ptr<Container> data)
+{
+  this->containers.push_back(std::move(data));
```
oruebel, line 16: By using `std::move` and using `std::unique_ptr<Container> data` we require that ownership is being transferred from the caller to to the ``containers`` vector.  I believe this means two things:
1. the caller cannot just call `addData` with a `Container` pointer must use `std::move` 
2. the caller loses ownership of the Container and cannot access the `data` anymore

E.g.:

```
RecordingContainers rc;
std::unique_ptr<Container> myContainer = std::make_unique<MyContainer>(1);
rc.addData(std::move(myContainer));  // Must use std::move here
// myContainer is a nullptr now
```

If we are saying that `RecordingContainers` is taking on ownership of the `Container`, then I think this makes sense. Does `NWBFile` need to keep the Container it creates around or does it give up ownership. I believe it is the former (in which case this is fine), but I just wanted to confirm.

However, this transfer of ownership is a detail that may not be obvious to users, so it would be useful to document this side-effect in the docstring of the function for developers and if a user is intended to call ``addData` directly (rather than, e.g.,`NWBFile.createElectricalSeries` calling it on the users behalf), then we should also mention it in the user workflow docs. However, if it is just happening behind the scences, then mentioning it in the docsting of ``addData`` should be sufficient.

tests/examples/testWorkflowExamples.cpp
```diff
@@ -0,0 +1,105 @@
...
+    // [example_workflow_recording_containers_snippet]
+
+    // [example_workflow_nwbfile_snippet]
+    std::unique_ptr<NWB::NWBFile> nwbfile =
+        std::make_unique<NWB::NWBFile>(generateUuid(), io);
+    nwbfile->initialize();
+    // [example_workflow_nwbfile_snippet]
+
+    // [example_workflow_datasets_snippet]
+    nwbfile->createElectricalSeries(
```
oruebel, line 51: How does the user know what the index of the `ElectricalSeries` is? I.e., how does the user know how to access the `ElectricalSeries`? Is this currently implicit, i.e., the user needs to remember the order in which they added objects to the `recordingContainers` object? 

If so, I'd suggest to either: 1) change `createElecticalSeries` to return the index of the object that was added, or 2) add a line here along the lines of ``SizeType electricalSeriesIndex = recordingContainers.get().size() - 1`` along with some text in the workflow docs to indicate that : a) new objects are always appended at the end in `RecordingContainers` and b) that this is how we can get and remember the index needed to retrieve the object later on. Either option seems fine, but it will be useful to document how we can get the index we should use to access a particular ``Container`` that was added to ``RecordingContainers``

docs/pages/userdocs/workflow.dox
```diff
@@ -0,0 +1,97 @@
...
+ * \subsection create_datasets 4. Create datasets and add to RecordingContainers.
+ *
+ * Next, create the different data types (e.g. \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries" 
+ * or other AQNWB::NWB::TimeSeries "TimeSeries") that you would like to write data into. After 
+ * creation, these objects are added to the  \ref AQNWB::NWB::RecordingContainers "RecordingContainers"  
+ * object so that it can manage access and data writing during the recording process. 
+ * 
+ * \snippet tests/examples/testWorkflowExamples.cpp example_workflow_datasets_snippet
+ *
+ *
```
oruebel, line 64: I would add a comment along the lines of 

*Ownership of the \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries" is being transferred to the \ref AQNWB::NWB::RecordingContainers "RecordingContainers" object, so that we can
again access it from here via ....* 

and then this is also where we could show how to retrieve the object `ElectricalSeries` object again and how to use the index.

oruebel (changes requested):
I added a couple of minor changes in `workflow.dox` (mainly to fix line breaks from my previous comments) and added a couple of minor comments to enhance documentation of a few details that may not be obvious to a user. Otherwise this looks good.

tests/examples/testWorkflowExamples.cpp
stephprince: I added `containerIndexes` as an optional input that will let users keep track of which index has been created by `createElectricalSeries`.

I think this approach will keep it more consistent that functions that might modify the NWBFile like `createElectricalSeries` return their status (e.g. returning a failure if the `io->canModifyObjects()` is false). 

`createElectricalSeries` can also add multiple containers (depending on the size of `recordingArrays`) to the `recordingContainers` object so I think having it within the function might also be easier for the user rather than keeping track of the size before and after `createElectricalSeries` is called.

oruebel (approved):

PR #86, opened 2024-09-02
Enable generation of CALL_GRAPH and CALLER_GRAPH in the docs

oruebel:
This PR enables the generation of caller and call graphs in Doxygen for all functions, providing an idea about who is calling the function and which functions are called by a function. E.g., here the call graph for ``createElectricalSeries``

Screen Shot 2024-09-01 at 7 53 05 PM

**Note**
* While this can increases the built time for the docs, on my laptop the difference seemed to be minimal
* Extracting the call/caller graph is not perfect in Doxygen. In particular in the case of virtual functions where we may call the function on the base type and not the derived type, callers seem to get missed at times. However, the graphs still seem useful, even if they are not perfect.


stephprince (approved):
Looks great!

PR #89, opened 2024-09-06
integration related updates

stephprince:
These are some minor updates related to the OpenEphys plugin integration [here](https://github.com/NeurodataWithoutBorders/nwb-format/tree/add-aqnwb).

- add data_collection and description as user provided inputs
- fix conversion of data from float to int (also some small changes to related tests)
- fix electrical series channel_conversion and electrodes attribute data write

### DIFF
--- a/src/Utils.hpp
+++ b/src/Utils.hpp
@@ -1,9 +1,13 @@
+#include <algorithm>
 #include <chrono>
+#include <cmath>
+#include <cstdint>
 #include <ctime>
 #include <iomanip>
 #include <sstream>
 
 #include <boost/date_time.hpp>
+#include <boost/endian/conversion.hpp>
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
@@ -68,6 +72,23 @@ inline std::shared_ptr<BaseIO> createIO(const std::string& type,
   }
 }
 
+inline void convertFloatToInt16LE(const float* source,
+                                  void* dest,
+                                  int numSamples)
+{
+  auto maxVal = static_cast<double>(0x7fff);
+  auto intData = static_cast<char*>(dest);
+
+  for (int i = 0; i < numSamples; ++i) {
+    auto clampedValue = std::clamp(maxVal * source[i], -maxVal, maxVal);
+    auto intValue =
+        static_cast<uint16_t>(static_cast<int16_t>(std::round(clampedValue)));
+    intValue = boost::endian::native_to_little(intValue);
+    *reinterpret_cast<uint16_t*>(intData) = intValue;
+    intData += 2;  // destBytesPerSample is always 2
+  }
+}
+
 inline std::unique_ptr<int16_t[]> transformToInt16(SizeType numSamples,
                                                    float conversion_factor,
                                                    const float* data)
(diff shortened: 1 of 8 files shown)


src/Utils.hpp
```diff
@@ -68,6 +72,23 @@ inline std::shared_ptr<BaseIO> createIO(const std::string& type,
   }
 }
 
+inline void convertFloatToInt16LE(const float* source,
+                                  void* dest,
+                                  int numSamples)
+{
+  auto maxVal = static_cast<double>(0x7fff);
+  auto intData = static_cast<char*>(dest);
```
oruebel, line 80: Can you clarify why dest is void* instead of char*? Since we use a static_cast to char* it seems that we assume that dest is always char*?

oruebel: I'm a bit confused. Why do we need to cast to char (i.e., an 8bit type) if intData is supposed be and int16?

src/Utils.hpp
```diff
@@ -68,6 +72,23 @@ inline std::shared_ptr<BaseIO> createIO(const std::string& type,
...
+
+  for (int i = 0; i < numSamples; ++i) {
+    auto clampedValue = std::clamp(maxVal * source[i], -maxVal, maxVal);
+    auto intValue =
+        static_cast<uint16_t>(static_cast<int16_t>(std::round(clampedValue)));
+    intValue = boost::endian::native_to_little(intValue);
+    *reinterpret_cast<uint16_t*>(intData) = intValue;
+    intData += 2;  // destBytesPerSample is always 2
```
oruebel, line 88: Is the data being converted to Int16 or Uint16? The function name implies Int16 but it seems the code is for UInt16. If it's UINT, then I would propose to change the name of the function accordingly. Also, please add a docstring for the function.

src/nwb/NWBFile.hpp
```diff
@@ -124,6 +127,9 @@ class NWBFile
   const std::string identifierText;
   std::shared_ptr<BaseIO> io;
   static std::vector<SizeType> emptyContainerIndexes;
+
+  std::string description;
+  std::string dataCollection;
```
oruebel, line 132: Since these are being written to the file, we should not save these on the NWBFile but instead make them accessible for read later.

oruebel (commented):
Just some minor questions/comments. Otherwise this looks good.

src/Utils.hpp
stephprince: Hmm good question, honestly I was just adapting this original [JUCE function](https://github.com/juce-framework/JUCE/blob/46c2a95905abffe41a7aa002c70fb30bd3b626ef/modules/juce_audio_basics/buffers/juce_AudioDataConverters.cpp#L41-L64) using the default values.

I believe the intermediate cast to char was for byte-wise pointer manipulation. Since we are just using the default value for destBytesPerSample, I'm not sure if we need this anymore. 

Similarly for the casting to uint16, we might be able to skip that step since we recast to int16 in the `transformToInt16` function anyway.

oruebel: What I would suggest then is to: 1) add a docstring to describe what the function is doing and 2) add some comments or TODO items in the function itself about these details so that if/when we need to revisit the function we can understand why these different types and casts are happening they way they are. As long as the function is working as intended, I don't think it is a high priority to change this right now.

stephprince: sounds good! Added

src/nwb/NWBFile.hpp
stephprince: update to not save on the NWBFile

oruebel (approved):

PR #91, opened 2024-09-08
Add read for neurodata_types, e.g., Container, TimeSeries

### LINKED ISSUE #103: Inconsistent use of paths
Different parts of the code assume path names with ending "/" and without ending "/". For example, `DynamicTable` assumes that it's path ends with "/"

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/file/ElectrodeTable.hpp#L89-L90

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.cpp#L45

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.cpp#L58

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.cpp#L71

However, other ``Containers`` assume that path names do not have a "/" at the end, e.g., here:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/base/TimeSeries.cpp#L51

To avoid errors and ease use of the API, I propose to:

1) Always use paths without ending "/"
2) In the constructor of ``Container`` remove trailing "/" from the path 
3) Update ``DynamicTable`` to not assume trailing "/"

oruebel:
This PR builds on #85. I made this a separate PR to make it a) easier to review and b) because this addresses the specific problem of reading a typed Container (e.g., a TimeSeries) and I wanted to make it easier to see the changes related to reading typed objects. Ultimately,  the goal of this PR is to enable us to read any object with an assigned neurodata_type (for which we have a corresponding class) from an NWB file. 

# How does read look like?

To read a type (whether a whole NWBFile or a single TimeSeries) , we need to know: 1) the full name of the type, 2) the path of the object in the file, and 3) the io object for reading. 

### Variant 1: User specifies name of type
```
auto readElectricalSeries = AQNWB::NWB::RegisteredType::create("core::ElectricalSeries", path, io);
```
### Variant 2: Users provides the type class as a template
```
auto readElectricalSeries =  NWB::RegisteredType::create<AQNWB::NWB::ElectricalSeries>(path, io);
```

or simply create the instance of the class ourselves: 

```
auto readElectricalSeries = AQNWB::NWB::ElectricalSeries(path, io);
```

### Variant 3: We read the full name of type from the file directly

```
auto readTS = AQNWB::NWB::RegisteredType::create(dataPath, io);
```

In this variant we read the full name of the type from the NWB file directly, which is a combination of the attributes ``namespace`` and ``neurodata_type`` that are stored for each type. 

# How does this work

I added more details about the design for this on a new page in the developer docs. 

# Main changes / TODO

- [X] Added``RegisteredType`` as common base class for all neurodata_types defined in the NWB (or HDMF) schema, i.e., for both Dataset and Group types
- [X] Updated all NWB/HDMF type classes to register with the ``RegisteredType`` registry
- [X] Updated ``Data``, ``VectorData``, and ``ElementIdentifiers`` to provide the standard constructor required by ``RegisteredType``. I.e., these types now also have the ``io`` and their path stored. 
- [X] Implemented class registry in ``RegisterType`` to manage all types based on the namespace and type-name
- [X] Implemented factory methods to create any registered type (e.g., TimeSeries) based on the name (e.g,. "core::TimeSeries"), io object, and path
- [X] Added unit tests for the ``RegisteredType`` class
- [X] Added developer docs on how to implement a subtype of ``RegisteredClass`` (i.e., a new data type class)
- [x]  Updated ``SpikeEventSeries``, e.g., to remove https://github.com/NeurodataWithoutBorders/aqnwb/blob/a6d86f99a600bc6d1777625fcf876970c70d32b9/src/nwb/ecephys/SpikeEventSeries.hpp#L60-L63
- [X] `ElectrodeTable` is not a type in NWB yet, as such we need to write with `DynamicTable` as the `neurodata_type` instead. Updated the `RegisteredType` class to allow manual overwrite of the neurodata_type when the classname is not the same as the neurodata_type.
- [X] Fix #103
    - [X] Added ``mergePaths`` function in ``Utils.hpp`` as a more reliable way to merge HDF5 paths   
    - [X] Updated all Container classes to use the new ``mergePaths` method and remove trailing "/" in path definitions
    - [X] Updated all helper functions in `Utils.hpp` to be `static inline` to avoid linker problem with multiple definitions of the functions and still be able to inline the functions (i.e. without having to move to a `Utils.cpp`
- [X] Added new macro `DEFINE_FIELD` in `RegisteredType` to simplify the creation of getter functions that return `ReadDataWrapper` object for lazy read of a dataset or attribute
    - [X] Updated Doxygen documentation settings to allow expansion of the DEFINE_FIELD macro to create documentation for the generated getter methods also in the docs 
- [X] Added methods to check whether a field actually exists 
    - [X] Added `ReadDataWrapper.exists` 
    - [X] Added `BaseIO::attributeExists` and `HDF5IO::attributeExists` (we already had ``BaseIO::objectExists`` for datasets/groups but not for attributes)
- [X] Add functionality to allow us to get dynamically defined typed groups (e.g,. to get `TimeSeries` from /acquisition)
   - [X] Add ``BaseIO::getGroupObjects`` and ``HDF5IO::getGroupObjects`` to allow us to get a list of all contents of a group
   - [X] Added ``BaseIO::findTypes`` to allow us to search for all objects of a given set of types. This function only uses functions defined on ``BaseIO`` so it is backend independent and does not need to be overridden in ``HDF5IO``
   - [X] Add example to show how to use ``findTypes`` on read
 - [x] Cleanup of naming convention in compliance with #102. Replace member variables with `m_` naming convention:
   - [x] ElectrodeTable
   - [x] ElectrodeGroup
   - [x] SpikeEventSeries
   - [x] ElectricalSeries
   - [x] Device
   - [x] DynamicTable
   - [x] VectorData
   - [x] Data
   - [X] TimeSeries 
   - [X] RegisteredType
   - [x] ReadIO
   - [x] DataBlock, DataBlockGeneric
   - [x] ReadDataWrapper classes
- [x] Update all Container classes to remove stored data values as attributes and replace with read getter functions to retrieve `ReadDataWrapper` objects for lazy read of datasets and attributes. These can be created with the new DEFINE_FIELD macro.
    - [x] TimeSeries
    - [x] ElectricalSeries
    - [x] SpikeEventSeries
    - [x] Device
    - [x] NWBFile
    - [x] ElectrodeTable
    - [x] DynamicTable
    - [x] Data
    - [x] VectorData
    - [ ] ... 
- [x] Support opening of a file in ``HDF5IO`` in read-only mode (probably using SWMR read)
- [ ] Implement read for compound data and references
- [ ] Add unit tests for reading
- [ ] Test that reading subsets of larger arrays works as expected
- [ ] Add not in developer test docs to clarify that tests on PR's are running on a temporary merge commit to test that the PR functions correctly when merged with the target branch (see https://github.com/NeurodataWithoutBorders/aqnwb/pull/85#issuecomment-2429899944)


oruebel:
@stephprince please take a look. This is an accompanying PR to #85 to implement data read, but this PR specifically focuses on read for typed objects (e.g., a ``TimeSeries``) where we need to construct a corresponding API class in AqNWB, instead of reading a single dataset or attribute. As far as I can tell this PR works, but it looks like there are still some bugs in reading attributes in #85.

oruebel:
@stephprince this PR is good to review.

tests/examples/test_ecephys_data_read.cpp
```diff
@@ -167,37 +168,36 @@ TEST_CASE("ElectricalSeriesReadExample", "[ecephys]")
...
-    io->stopRecording();
-    io->close();
-    // [example_read_finish_recording_snippet]
-
-    // [example_read_only_snippet]
     // Open an I/O for reading
-    std::shared_ptr<BaseIO> readio = createIO("HDF5", path);
+    // TODO: creating a new I/O makes the read fail.
+    // std::shared_ptr<BaseIO> readio = createIO("HDF5", path);
+    // readio->open();
```
oruebel, line 174: @stephprince it looks like there is an issue with how HDF5IO opens an existing file. Read works fine when I use the existing HDF5IO object that was used for write, but when I close the file and then open it again reading any data from the file fails. When I tried this in the  ``testBase.cpp`` it corrupted the file (i.e,. I think it may be opened in "write" mode  instead of "read" or "append").

oruebel: @stephprince this may be related to the issue you encountered on Windows

oruebel:
@stephprince just FYI. I pushed a few more updates here. The main structural changes are: 1) I created a ``DEFINE_FIELD`` macro to make it easier to create standardized getter methods for reading datasets/attributes and 2) I added a ``mergePaths`` method to standardize the handling of paths a bit more. 

Next, I'll need to:
* add all the `DEFINE_FIELD` definitions to allow us to read all the fields 
* add methods to allow us to retrieve the paths for dynamically created Groups (e.g., ``ElectricalSeries``) to make it easier to also get to the objects that are not at fixed locations in the file. As long as a user knows the paths, we can already open these, but we don't yet have a way for the user to find the paths.
* Once that is done, we'll need to add a lot of tests. I'm sure there is still some debugging ahead to make sure reading all the different types of data and slicing works as expected.

oruebel:
> * add methods to allow us to retrieve the paths for dynamically created Groups (e.g., `ElectricalSeries`) to make it easier to also get to the objects that are not at fixed locations in the file. As long as a user knows the paths, we can already open these, but we don't yet have a way for the user to find the paths.

@stephprince I implemented this now as follows:

- Added ``BaseIO::getGroupObjects`` and ``HDF5IO::getGroupObjects`` to allow us to get a list of all contents of a group
- Added ``BaseIO::findTypes`` to allow us to search for all objects of a given set of types. This function only uses functions defined on BaseIO so it is backend independent and does not need to be overridden in HDF5IO
- Add example to show how to use ``findTypes`` on read

oruebel:
@stephprince if it's easier for you to review, I'd be fine with merging this back to #85

stephprince:
> I pushed a few more updates here. The main structural changes are: 1) I created a DEFINE_FIELD macro to make it easier to create standardized getter methods for reading datasets/attributes and 2) I added a mergePaths method to standardize the handling of paths a bit more.

> Added BaseIO::getGroupObjects and HDF5IO::getGroupObjects to allow us to get a list of all contents of a group
> Added BaseIO::findTypes to allow us to search for all objects of a given set of types. This function only uses functions defined on BaseIO so it is backend independent and does not need to be overridden in HDF5IO
> Added example to show how to use findTypes on read

@oruebel sounds great!

> if it's easier for you to review, I'd be fine with merging this back to https://github.com/NeurodataWithoutBorders/aqnwb/pull/85

Can we keep it separate for now? I think I had read most of the docs and examples for this one and I want to keep track of where I left off.

oruebel:
> Can we keep it separate for now? I think I had read most of the docs and examples for this one and I want to keep track of where I left off.

Sounds good. Whichever works best for you is fine with me.

stephprince (commented):
Overall looks good and the docs explained things very well, mainly some very small suggestions for clarity.

I still need to look into the issue with `HDF5IO` opening an existing file.

Also, once the small suggestions are updated, if it's easier to combine the two read PRs I'm good with that.

tests/examples/test_ecephys_data_read.cpp
stephprince: I am having trouble reproducing this on the main branch. 

If I add these lines to the end of `testBase.cpp` I am able to create a new IO object and read the data back in. 

```c++
    io->close();

    // test reopening the file to read the data back in
    std::shared_ptr<BaseIO> readio = createIO("HDF5", path);
    readio->open();
    double* tsBuffer2 = new double[numSamples];
    std::unique_ptr<BaseRecordingData> tsetDset2 =
        readio->getDataSet(dataPath + "/timestamps");
    std::unique_ptr<HDF5::HDF5RecordingData> tsH5Dataset2(
        dynamic_cast<HDF5::HDF5RecordingData*>(tsetDset2.release()));
    readH5DataBlock(tsH5Dataset2->getDataSet(), timestampsType, tsBuffer2);
    std::vector<double> tsRead2(tsBuffer2, tsBuffer2 + numSamples);
    delete[] tsBuffer2;
    REQUIRE(tsRead2 == timestamps);
```

Do you remember if there were any different steps or conditions?

docs/pages/devdocs/registered_types.dox
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * 2. **Retrieving Registered Subclass Names**:
+ *    - Use the \ref AQNWB::NWB::RegisteredType::getRegistry "getRegistry" method to retrieve the
+ *      set of registered subclass names.
+ *      \snippet tests/examples/test_RegisteredType_example.cpp example_RegisterType_get_registered_names
+ *
+ * 3. **Retrieving the Factory Map**:
+ *    - Use the \ref AQNWB::NWB::RegisteredType::getFactoryMap "getFactoryMap" method to retrieve
+ *      the map of factory functions for creating instances of registered subclasses.
+ *      \snippet tests/examples/test_RegisteredType_example.cpp example_RegisterType_get_registered_factories
```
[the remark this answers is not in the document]
oruebel: I don't think a user would typically call these functions directly. I think a user would maybe call ``RegisteredType::create``, but I think even for that we will probably ultimately have some other access methods that would call it on the user's behalf . E.g., if we want to read an ``ElectricalSeries``, I think we probably would want some method on ``NWBFile`` to get `TimeSeries` objects from the `acquisition` group.

src/Utils.hpp
```diff
@@ -75,6 +75,48 @@ inline std::shared_ptr<IO::BaseIO> createIO(const std::string& type,
...
+ * This function is useful for constructing paths in HDF5 files, where paths
+ * always use "/".
+ *
+ * @param path1 The first path to merge.
+ * @param path2 The second path to merge.
+ *
+ * @return The merged path with no extra "/" at the start or end, and exactly
+ * one "/" between the two input paths.
+ */
+static inline std::string mergePaths(const std::string& path1,
```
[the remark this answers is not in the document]
oruebel: I have not used ``std::filesystem`` a lot, but I think the answer is yes. One issue to consider is that ``std::filestystem`` I believe automatically decides on whether to use "/" or "\" in paths, depending on system (i.e., Windows vs Unix). However, for HDF5 we always want to use path's with "/". I believe we can force `std::filesystem` to use "/", so just something to keep in mind. I we want to use `std::filesystem` for this, then I would maybe deal with that as a separate issue.

tests/examples/test_ecephys_data_read.cpp
oruebel: Thanks for confirming. A simple test is to just uncomment the two lines here where the `readio` object is being created and then use the `readio` object instead of the original `io` object in the code that follows. We'd also need to call `io.close` first since we don't have a way yet to open as read-only with SWMR-read.

docs/pages/userdocs/read.dox
```diff
@@ -219,23 +223,151 @@
...
  *
+ * \subsection read_design_sec_read_types Reading typed objects
+ *
+ * Objects with an assigned ```neurodata_type`` are represented by corresponding classes in AqNWB.
+ * To read objects with an assigned type, we therefore need to be able to instantiate the corresponding
+ * classes in AqNWB based on the data from a file. The following figure illustrates the main
+ * components of this process.
+ *
+ *
+ * @dot
```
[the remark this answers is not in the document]
oruebel: Done

docs/pages/userdocs/read.dox
```diff
@@ -219,23 +223,151 @@
...
+ * The main components involved in reading typed objects from an NWB file via AqNWB are:
+ *
+ * - \ref AQNWB::NWB::RegisteredType "RegisteredType" as the main base class for all classes
+ *   implementing a type, e.g., \ref AQNWB::NWB::Container "Container" \ref AQNWB::NWB::Data "Data"
+ *   and all their subtypes. \ref AQNWB::NWB::RegisteredType "RegisteredType" is responsible for
+ *   managing all type classes and provides the \ref AQNWB::NWB::RegisteredType::create "create"
+ *   factory methods for creating instances of subclasses from a file.
+ * - \ref AQNWB::IO::BaseIO "BaseIO", \ref AQNWB::IO::HDF5::HDF5IO "HDF5IO" responsible for
+ *   reading type attribute and group information and for searching the file for typed objects via
+ *   \ref AQNWB::IO::BaseIO::findTypes "findTypes()", and exposing
```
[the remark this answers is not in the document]
oruebel: Fixed

oruebel:
> Also, once the small suggestions are updated, if it's easier to combine the two read PRs I'm good with that.

@stephprince I added the changes you suggested and synced the PR with the `add_read` branch.

tests/examples/test_ecephys_data_read.cpp
stephprince: I'm not positive since there were some code changes since these comments, but I think the main issue was that `readDataWrapper` was being used here instead of `readElectricalSeriesData` obtained from the `read_io` object.

```c++
auto readElectricalSeriesData = readElectricalSeries->dataLazy();
DataBlock<float> readDataValues = readDataWrapper->values<float>();
```

This test example now works using the `read_io` object, but the file is being re-opened in read/write mode. I will add an option to open in read-only mode and demonstrate that here.

oruebel: > I think the main issue was that `readDataWrapper` was being used here instead of `readElectricalSeriesData`

Makes sense. Thanks for catching that mistake. 

> I will add an option to open in read-only mode and demonstrate that here.

For read-only mode we should also allow SWMR read mode.

stephprince:
To focus on getting the initial read functionality merged into the main branch - I propose we now:
- merge this branch with #85 and
- bump the following items to follow up PRs
    - [ ] Implement read for compound data and references
    - [ ] Add not in developer test docs to clarify that tests on PR's are running on a temporary merge commit to test that the PR functions correctly when merged with the target branch (see https://github.com/NeurodataWithoutBorders/aqnwb/pull/85#issuecomment-2429899944)

oruebel:
> To focus on getting the initial read functionality merged into the main branch - I propose we

That all sounds good to me. 

> I updated the Container classes with read getter functions using `DEFINE_FIELD`. I may have missed removing/adding some things so a review of that would be helpful.

Thanks for all the updates. I'll take a look at the commits you added this/last week.

stephprince:
> Thanks for all the updates. I'll take a look at the commits you added this/last week.

Sounds good! Will squashing and merging this branch into the `add_read` branch work? Or would you prefer I wait for you to review.

src/nwb/NWBFile.cpp
```diff
@@ -28,22 +28,31 @@ constexpr SizeType SPIKE_CHUNK_XSIZE =
...
+
 NWBFile::~NWBFile() {}
 
 Status NWBFile::initialize(const std::string& identifierText,
                            const std::string& description,
                            const std::string& dataCollection)
 {
   if (std::filesystem::exists(m_io->getFileName())) {
-    return m_io->open(false);
+    return m_io->open(IO::FileMode::ReadWrite);
```
oruebel, line 53:
```suggestion
    return m_io->open();
```

I don't think we need to specify the mode here, since the `open()` without parameters does this for us

src/nwb/NWBFile.cpp
```diff
@@ -28,22 +28,31 @@ constexpr SizeType SPIKE_CHUNK_XSIZE =
...
   } else {
-    m_io->open(true);
+    m_io->open(IO::FileMode::Overwrite);
```
oruebel, line 55:
```suggestion
    m_io->open();
```

Same as above. I don't think we need to specify the mode here, since the `open()` without parameters does this for us.

oruebel:
> Or would you prefer I wait for you to review.

Give me another 30min to review. I'll ping you when I'm done.

src/nwb/ecephys/ElectricalSeries.hpp
```diff
@@ -83,15 +88,25 @@ class ElectricalSeries : public TimeSeries
...
+               Channel - specific conversion factor)
 
+  DEFINE_FIELD(readData, DatasetField, std::any, "data", Recorded voltage data)
+
+  DEFINE_FIELD(readDataUnit,
+               AttributeField,
+               std::string,
+               "data/unit",
+               Base unit of measurement for working with the data. 
+               This value is fixed to volts)
```
oruebel, line 104: `readData` and `readDataUnit` are already specified on TimeSeries. I think this should work, but just something to double-check to make sure DEFINE_FIELD successfully overwrite those inherited functions.

oruebel:
The changes you made look all good to me. I added a couple of minor comments. Feel free to merge with the other branch.

src/nwb/NWBFile.cpp
stephprince: thanks for catching! I missed that the file mode checking was duplicated here from `open()`

src/nwb/ecephys/ElectricalSeries.hpp
stephprince: The documentation displays the `readData` and `readDataUnit` descriptions as added in ElectricalSeries instead of the descriptions from the inherited TimeSeries functions. I think that suggests that DEFINE_FIELD successfully overwrote those functions, but we can definitely double check with the unit tests as well.

oruebel: Yes, double-checking with the unit tests will be useful here as the Doxygen may behave different than the actual code. To get `DEFINE_FIELD` to work with Doxygen, I had to create it is own (simplified) definition in the `Doxyfile.in` : 

https://github.com/NeurodataWithoutBorders/aqnwb/blob/6a7bb45d5f280c300763b8427a958acda6be2dfc/docs/Doxyfile.in#L10-L19

I.e., Doxygen is not using `[RegisteredType::DEFINE_FIELD](https://github.com/NeurodataWithoutBorders/aqnwb/blob/6a7bb45d5f280c300763b8427a958acda6be2dfc/src/nwb/RegisteredType.hpp#L288-L325)` when creating the documentation, so the behavior may be different. This is not ideal, but was the only way that seemed to work with Doxygen (everything else I tried to get Doxygen to actually recognize the real macro failed).

I believe the code is working correctly, but it will still be good to verify this in the unit tests.

Issue #30, opened 2024-06-05
add SpikeEventSeries data type


stephprince:
Add additional flexibility to startRecording method to record the SpikeEventSeries datatype

stephprince:
This was addressed by https://github.com/NeurodataWithoutBorders/aqnwb/pull/92

Issue #37, opened 2024-07-26
make ChannelVector a class


stephprince:
Currently ChannelVector is an alias for a vector of channels which are used to group together channel-related information from the user / acquisition system. We would like to save additional information about the channel group (e.g. group name, comments, tags, etc.). See discussion here for more information: https://github.com/NeurodataWithoutBorders/aqnwb/pull/33#discussion_r1692309745 

Also we should discuss with @rly an appropriate name given the ongoing restructuring to ElectrodeTable.

PR #92, opened 2024-09-09
Add SpikeEventSeries data type

stephprince:
These changes add the `SpikeEventSeries` data type for recording threshold crossings during data acquisition. In addition to adding the new data type, these changes check whether the `ElectrodesTable` already exists on the `NWBFile`. (Note: what I've written so far assumes that once the electrodes table has been created, all electrodes have been added. I think the `ElectrodesTable` / `DynamicTable` classes would require further changes if we want to append to an existing table).

I think some additional tests are needed to check that the code :
- [x] succeeds for different dimensions of the SpikeEventSeries ([nEvents x nSamples], [nEvents x nChannels x nSamples])
- [x] succeeds if user makes file with both spike event series + electrical series


src/BaseIO.hpp
```diff
@@ -292,6 +292,13 @@ class BaseIO
   virtual std::unique_ptr<BaseRecordingData> getDataSet(
       const std::string& path) = 0;
 
+  /**
+   * @brief Checks whether an object already exists at the location in the file.
```
oruebel, line 296:
```suggestion
   * @brief Checks whether a Dataset, Group, or Link already exists at the location in the file.
```

``H5Lexists`` only works for Datasets and Groups (but not Attributes). I don't think this function needs to cover attributes right now, but I figured it would be useful to clarify this in the docstring.

src/hdf5/HDF5IO.hpp
```diff
@@ -234,6 +234,13 @@ class HDF5IO : public BaseIO
   std::unique_ptr<BaseRecordingData> getDataSet(
       const std::string& path) override;
 
+  /**
+   * @brief Checks whether an object already exists at the location in the file.
```
oruebel, line 238:
```suggestion
   * @brief Checks whether a Dataset, Group, or Link already exists at the location in the file.
```

src/nwb/NWBFile.cpp
```diff
@@ -142,14 +146,88 @@ Status NWBFile::createElectricalSeries(
...
+    // Setup Spike Event Series datasets
+    SizeArray dsetSize;
+    SizeArray chunkSize;
+    if (channelVector.size() == 1) {
+      dsetSize = SizeArray {0, 0};
+      chunkSize = SizeArray {SPIKE_CHUNK_XSIZE, 1};
+    } else {
+      dsetSize = SizeArray {0, channelVector.size(), numSamples};
+      chunkSize = SizeArray {SPIKE_CHUNK_XSIZE, 1, 1};
+    }
```
oruebel, line 207: I think we here also need to check that the ``ElectrodeGroup`` and ``Device`` do not already exist. When recording a `SpikeEventSeries`, chances are we are also recording an `ElectricalSeries` with the raw data from the same device and channel at the same time. One option may be to have a variant of this function that would take a ``std::vector<ElectrodeGroup>`` instead of ``std::vector<Types::ChannelVector> recordingArrays`` as input. The same is true for ``createElectricalSeries``, depending on the order in which the user creates different TimeSeries for the same group of channels.

src/nwb/NWBFile.hpp
```diff
@@ -127,9 +148,11 @@ class NWBFile
       const std::array<std::pair<std::string_view, std::string_view>, N>&
           specVariables);
 
+  std::unique_ptr<ElectrodeTable> elecTable;
   const std::string identifierText;
   std::shared_ptr<BaseIO> io;
   static std::vector<SizeType> emptyContainerIndexes;
+  inline const static std::string acquisitionPath = "/acquisition/";
```
oruebel, line 155:
```suggestion
  inline const static std::string acquisitionPath = "/acquisition";
```

If I remember correctly, in other places, paths are defined without the "/" at the end. It would be useful to be consistent to make it easier to merge paths without having to check whether we already have "/" at the end.  We could consider switching to using ``std::filesystem::path`` for path strings (we just need to remember to call ``generic_string()`` to construct the path to avoid having "\" instead of "/" on Windows, because HDF5 I believe always wants paths with "/")

src/nwb/ecephys/SpikeEventSeries.hpp
```diff
@@ -0,0 +1,70 @@
...
+  Status writeSpike(const SizeType& numSamples,
+                    const SizeType& numChannels,
+                    const void* data,
+                    const void* timestamps);
+
+private:
+  /**
+   * @brief The neurodataType of the SpikeEventSeries.
+   */
+  std::string neurodataType = "SpikeEventSeries";
```
oruebel, line 63: Note, with #91 we won't need to store the neurodataType, because each of these classes is being registered and will automatically get a function to get the name and namespace for the type.

oruebel (changes requested):
Overall this looks good. The main question I have is how we can prevent duplication of 1)  channels in the ``ElectrodeTable``, 2) ``ElectrodeGroup``s , 3) ``Device``s

src/nwb/NWBFile.cpp
stephprince: I added a check whether the device exists for a specific channel group since in our current workflow the `Device` and `ElectrodeGroup` for each channel group always created at the same time. 

More long-term, I'm wondering if we might want to incorporate this to be a part of the `Container` constructor: check if the group / dataset exists already, create it if it doesn't exist, otherwise read it in? I'm not sure if that leads to more complications in other ways that I'm not thinking of....

src/nwb/ecephys/SpikeEventSeries.hpp
stephprince: Sounds great!

src/nwb/NWBFile.hpp
stephprince: Since HDF5 always expects "/" for paths and I don't think there's anywhere we're doing any path manipulation, I don't know if it's worth it to switch to `std::filesystem::path` right now. But I could see maybe it would become more useful to have the path objects when you're trying to determine the contents of groups and the file?

stephprince:
For duplication of channels in the `ElectrodesTable`, my first thought is to:
1.  In `ElectrodeTable->addElectrodes()`, add a check using the global index/rowIds to only add channels not yet included 
2. In `ElectrodeTable->finalize()`, if the columns already exist, use an `addRow` method to append additional channels, otherwise create the columns with all the data.

Thoughts on that approach?

src/nwb/NWBFile.cpp
```diff
@@ -107,27 +110,33 @@ Status NWBFile::createElectricalSeries(
...
+  }
 
   // Create continuous datasets
   for (const auto& channelVector : recordingArrays) {
     // Setup electrodes and devices
     std::string groupName = channelVector[0].groupName;
-    std::string devicePath = "/general/devices/" + groupName;
-    std::string electrodePath = "/general/extracellular_ephys/" + groupName;
-    std::string electricalSeriesPath = rootPath + groupName;
+    std::string sourceName = channelVector[0].sourceName;
```
stephprince, line 125: One thing I realized when adding tests for a file that has both `ElectricalSeries` and `SpikeEventSeries` was that we needed to have a way to specify that data came from the same source (e.g. the same device) but had different names within the acquisition group. I added this `sourceName` as a separate input to the `Channel` class as a way to accommodate this.

stephprince:
Added tests and now also have it integrated with OpenEphys - see [here](https://github.com/NeurodataWithoutBorders/nwb-format/tree/add-aqnwb) for the latest version of that branch.

oruebel:
> For duplication of channels in the `ElectrodesTable`, my first thought is to:
> 
>     1. In `ElectrodeTable->addElectrodes()`, add a check using the global index/rowIds to only add channels not yet included
> 
>     2. In `ElectrodeTable->finalize()`, if the columns already exist, use an `addRow` method to append additional channels, otherwise create the columns with all the data.
> 
> 
> Thoughts on that approach?

Seems reasonable. I think 2. assumes that there are no custom columns added to the table. We may want to at least check that we have data for all columns

src/nwb/NWBFile.cpp
oruebel: > More long-term, I'm wondering if we might want to incorporate this to be a part of the `Container` constructor: check if the group / dataset exists already, create it if it doesn't exist, otherwise read it in? I'm not sure if that leads to more complications in other ways that I'm not thinking of....

Seems reasonable, but I believe the check for whether the object exists or not should appear in the ``initialize`` method instead of the constructor. The constructor is responsible for creating the instance, but should not do any real work on write (nor read). The main difference between read and write is not how the Container object is being constructed, but what functions are called afterwards, i.e., on write we call ``initialize`` and on read the user will call methods to get lazy read access to properties of the ``Container``.

src/nwb/NWBFile.cpp
oruebel: Is ``sourceName`` the same as the name of the ``device``?

stephprince: This is now updated so that `groupName` is used for the ElectrodeGroup and Device and is pulled from the `channelVector` input. The name of the dataset within the `acquisition` group is set by a `recordingArrayNames` input to `createElectricalSeries` and `createSpikeEventSeries`.

stephprince:
@oruebel this is ready for you to review again. I've created issues for what we want to address outside this PR

stephprince:
Will fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/30

src/nwb/NWBFile.cpp
```diff
@@ -107,27 +111,44 @@ Status NWBFile::createElectricalSeries(
...
+    const std::string& recordingName = recordingNames[i];
 
-  // Create continuous datasets
-  for (const auto& channelVector : recordingArrays) {
     // Setup electrodes and devices
     std::string groupName = channelVector[0].groupName;
     std::string devicePath = "/general/devices/" + groupName;
     std::string electrodePath = "/general/extracellular_ephys/" + groupName;
-    std::string electricalSeriesPath = rootPath + groupName;
+    std::string electricalSeriesPath = acquisitionPath + "/" + recordingName;
```
oruebel, line 140: For now this seems fine. I assume this is part of https://github.com/NeurodataWithoutBorders/aqnwb/issues/93 to address the groupName?

stephprince: By address the `groupName` do you mean allowing different options for device and electrodeGroup?

From what I had in my notes we also discussed having a single `groupName` for a channelVector. That would be part of https://github.com/NeurodataWithoutBorders/aqnwb/issues/37

oruebel: Makes sense.

oruebel (approved):

Issue #15, opened 2024-04-01
Add CI for Windows


oruebel:
See also https://github.com/lbl-cbg/aq-nwb/pull/12

PR #99, opened 2024-09-17
update ci to include windows

stephprince:
Fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/15.

These changes add support for windows to our GitHub Actions workflows. As part of these changes, I addressed some build errors and warnings:
- linked bcrypt in the `CMakeLists.txt` on windows for `boost-uuid` functions. In think usually this should be auto-linked, but installing `boost` with `vcpkg` seems to disable this auto-linking.
- split up long spec strings (specifically the `nwb_file` one) into multiple components to avoid strings being too long for the compiler. I'm not sure how conservative we want to be on the maximum string length
- updated the SWMR tests
   - used a platform-independent sleep function
   - used conditional compiling for calling the reader executable
- fixed various small warnings (unused variables, shadowed variables, implicit casting)

Note: squash when merging


stephprince:
Everything is now be building/installing successfully and the tests are running. 

However, the SWMR mode tests are failing on windows. I'm not sure what exactly might be causing these differences on the windows tests vs. the macOS and ubuntu tests, whether it is an issue with the SWMR mode setup or some other file permission issue. @oruebel do you have any ideas?

oruebel:
Not sure. Windows tends to lock file access more strongly so maybe the file isn't being closed correctly or opened in non-swmr read mode by the reader. There is a separate read mode for SWMR if I remember correctly.

oruebel:
@stephprince I synced the PR with the main branch. Just wanted to mention just in case, since there were a number of merge conflicts, but I believe I resolved them all [ffffdc5](https://github.com/NeurodataWithoutBorders/aqnwb/pull/99/commits/ffffdc513fab8f2afa15e203a33d0e1d7d2656a6) correctly

oruebel:
> I believe I resolved them all [ffffdc5](https://github.com/NeurodataWithoutBorders/aqnwb/pull/99/commits/ffffdc513fab8f2afa15e203a33d0e1d7d2656a6) correctly

Spoken too soon. But I updated the tests and the build is working now. There is a failure on the tests, but I think that was already happening before the update.

stephprince:
@oruebel I did some further testing on the windows machine. From what I can tell, the reader process is opening the file using the SWMR mode read mode, but this process is preventing write access due to file locking. It appears that there are known issues that SWMR mode on Windows is not as well tested and does not always work:  https://github.com/h5py/h5py/issues/2022.

I can set the environment variable `HDF5_USE_FILE_LOCKING=FALSE` in the reader thread of the tests specifically on windows as a workaround. However since removing file locking could expose a user to other problems with file access, it may be safer for general usage to not recommend the use of SWMR mode on Windows.  I believe SWMR read mode is working as expected on macOS/ubuntu.

oruebel (dismissed):

oruebel:
> I can set the environment variable `HDF5_USE_FILE_LOCKING=FALSE` in the reader thread of the tests specifically on windows as a workaround.

For the unit tests that seems reasonable.

> However since removing file locking could expose a user to other problems with file access, it may be safer for general usage to not recommend the use of SWMR mode with AqNWB on Windows. I believe SWMR read mode is working as expected on macOS/ubuntu.

I assume the issue should only arise when trying to open the file again for read while it is still being written to. In that case, I think using SWMR mode for write should still be fine. I would add a warning to the SWMR tutorial to mention the Windows issues and the work-around. We could even added a static convenience method on HDF5IO to "disableFileLocking". As long as the user has to make the explicit choice to do so, I think this is fine.

stephprince:
@oruebel I updated the documentation to add a warning - can you reapprove this PR to merge

oruebel (approved):

PR #102, opened 2024-09-19
Update member variable names to avoid shadowing

### LINKED ISSUE #9: Adopt naming convention for member variables
Use `this` (e.g. `this->variableName` for member variables) or prefixes (e.g.`mVariableName` for member and `sVariableName` for static variables) to distinguish local and member variables.

### LINKED ISSUE #60: Shadowing of variables
Variables are being shadowed because function arguments can class members use the same naming convention. E.g.,

https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/Channel.hpp#L56-L59

defines the member variable `name` and the constructor also defined `name` as constructor argument:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/Channel.hpp#L21

this results in 

https://github.com/NeurodataWithoutBorders/aqnwb/blob/00aa1caff04ce9938b4d07c937859cdc34004539/src/Channel.cpp#L16-L24

where the object member `name` is being initialize with the constructor argument `name`. This results in a lot of compiler warnings and can also be confusing to read. 

Related to #9

### LINKED ISSUE #106: Hiden inherited member
In a couple of places, member variables inherited from the base class are being re-declared in the child class, hiding the inherited member and creating a duplicate member.  I believe this is unintended behavior and can lead to strange and unexpected behavior, as we now have two members with the same name in the same object. 

# What is the problem:

```
class A:
   std::string my_name;

class B : public A
   std::string my_name;
```

In C++, when a derived class ``B`` inherits from a base class ``A``, it inherits all the members of the base class, including all data members like ``my_name``. However, when declareing another member with the same name ``my_nam`` in the derived class ``B``, it hides the inherited member from the base class.

So, in this case, class ``B`` actually has two instances of ``my_name``: 1) one inherited from class A and 2) one declared in class B itself. However, the inherited ``my_name`` from class ``A`` is hidden by the ``my_name`` declared in class ``B``. However, one can still access the hidden ``my_name`` from class ``A`` by defining the scope of resolution operator ``A::my_name`` in class ``B``. 

This also means that when setting ``my_name``, only `B::my_name`` is being set but `A::my_name` remains unset. This is a problem, since any functions inherited from A may only see `A::my_name`

# Where is this happening

`ElectrodeTable.colNames` is hiding `DynamicTable.colNames`:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.hpp#L91-L94

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/file/ElectrodeTable.hpp#L118-L121

`HDF5IO.fileName` is hiding `BaseIO.fileName`:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/BaseIO.hpp#L349-L352

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/hdf5/HDF5IO.hpp#L267

oruebel:
Fix #60
Fix #9
Fix #106

- [X] Container
- [X] Data
- [X] NWBFile
- [X] RecordingContainers
- [X] Channel
- [X] VectorData  
- [X] ElementIdentifiers (noop)
- [x] HDF5RecordingData
- [x] HDF5IO
- [x] BaseIO
- [x] DynamicTable

**To be fixed in read I/O PR #85**
- [ ] ElectrodeTable 
- [ ] ElectrodeGroup
- [ ] SpikeEventSeries
- [ ] ElectricalSeries
- [ ] Device
- [ ] TimeSeries (already fixed in #85)

### DIFF
--- a/src/BaseIO.cpp
+++ b/src/BaseIO.cpp
@@ -31,9 +31,10 @@ const BaseDataType BaseDataType::DSTR = BaseDataType(T_STR, DEFAULT_STR_SIZE);
 
 // BaseIO
 
-BaseIO::BaseIO()
-    : readyToOpen(true)
-    , opened(false)
+BaseIO::BaseIO(const std::string& filename)
+    : m_filename(filename)
+    , m_readyToOpen(true)
+    , m_opened(false)
 {
 }
 
@@ -41,12 +42,17 @@ BaseIO::~BaseIO() {}
 
 bool BaseIO::isOpen() const
 {
-  return opened;
+  return m_opened;
+}
+
+std::string BaseIO::getFileName()
+{
+  return this->m_filename;
 }
 
 bool BaseIO::isReadyToOpen() const
 {
-  return readyToOpen;
+  return m_readyToOpen;
 }
 
 bool BaseIO::canModifyObjects()
--- a/src/BaseIO.cpp
+++ b/src/BaseIO.cpp
@@ -47,7 +47,7 @@ bool BaseIO::isOpen() const
 
 std::string BaseIO::getFileName()
 {
-  return this->m_filename;
+  return m_filename;
 }
 
 bool BaseIO::isReadyToOpen() const
--- a/src/BaseIO.cpp
+++ b/src/BaseIO.cpp
@@ -40,26 +40,6 @@ BaseIO::BaseIO(const std::string& filename)
 
 BaseIO::~BaseIO() {}
 
-bool BaseIO::isOpen() const
-{
-  return m_opened;
-}
-
-std::string BaseIO::getFileName()
-{
-  return m_filename;
-}
-
-bool BaseIO::isReadyToOpen() const
-{
-  return m_readyToOpen;
-}
-
-bool BaseIO::canModifyObjects()
-{
-  return true;
-}
-
 Status BaseIO::createCommonNWBAttributes(const std::string& path,
                                          const std::string& objectNamespace,
                                          const std::string& neurodataType,
--- a/src/Channel.hpp
+++ b/src/Channel.hpp
@@ -39,142 +39,166 @@ public:
   /**
    * @brief Copy constructor
    */
-   Channel(const Channel& other) = default;
+  Channel(const Channel& other) = default;
 
-   /**
-    * @brief Move constructor
-    */
-   Channel(Channel&& other) = default;
+  /**
+   * @brief Move constructor
+   */
+  Channel(Channel&& other) = default;
 
-   /**
-    * @brief Assignment operator
-    */
-   Channel& operator=(const Channel& other) = default;
+  /**
+   * @brief Assignment operator
+   */
+  Channel& operator=(const Channel& other) = default;
 
-   /**
-    * @brief Move assignment operator
-    */
-   Channel& operator=(Channel&& other) = default;
+  /**
+   * @brief Move assignment operator
+   */
+  Channel& operator=(Channel&& other) = default;
 
-   /**
+  /**
    * @brief Getter for conversion factor
    * @return The conversion value.
    */
   inline float getConversion() const { return m_bitVolts / m_conversion; }
 
-   /**
+  /**
    * @brief Getter for sampling rate of the channel.
    * @return The sampling rate value.
    */
-   inline float getSamplingRate() const { return m_samplingRate; }
+  inline float getSamplingRate() const { return m_samplingRate; }
 
-   /**
+  /**
    * @brief Getter for bitVolts floating point value of microvolts per bit
    * @return The bitVolts value.
    */
-   inline float getBitVolts() const { return m_bitVolts; }
+  inline float getBitVolts() const { return m_bitVolts; }
 
-   /**
+  /**
    * @brief Get the name of the array group the channel belongs to.
    * @return The groupName value.
    */
-   inline std::string getGroupName() const {  return m_groupName; }
+  inline std::string getGroupName() const { return m_groupName; }
 
-   /**
+  /**
    * @brief Get the name of the channel
    * @return The name value.
    */
-   inline std::string getName() const { return m_name; }
+  inline std::string getName() const { return m_name; }
 
-   /**
+  /**
    * @brief Get the array group index the channel belongs to
    * @return The groupIndex value.
    */
-   inline SizeType getGroupIndex() const { return m_groupIndex; }
+  inline SizeType getGroupIndex() const { return m_groupIndex; }
 
-   /**
+  /**
    * @brief Get the index of the channel within the recording array.
    * @return The localIndex value.
    */
-   inline SizeType getLocalIndex() const { return m_localIndex; }
+  inline SizeType getLocalIndex() const { return m_localIndex; }
 
-   /**
+  /**
    * @brief Get the index of the channel across the recording system.
    * @return The globalIndex value.
    */
-   inline SizeType getGlobalIndex() const { return m_globalIndex; }
+  inline SizeType getGlobalIndex() const { return m_globalIndex; }
 
-   /**
+  /**
    * @brief Get the coordinates of channel (x, y, z) within the recording array.
    * @return The position value.
    */
-   inline const std::array<float, 3>& getPosition() const { return m_position; }
+  inline const std::array<float, 3>& getPosition() const { return m_position; }
 
-   /**
+  /**
    * @brief Get comments about the channel
    * @return The comments value.
    */
-   inline std::string getComments() const { return m_comments; }
+  inline std::string getComments() const { return m_comments; }
 
-   /**
+  /**
    * @brief Set comments about the channel.
    * @param comments The comments to set.
    */
-   inline void setComments(const std::string& comments) { m_comments = comments; }
+  inline void setComments(const std::string& comments)
+  {
+    m_comments = comments;
+  }
 
-   /**
+  /**
    * @brief Set coordinates of channel (x, y, z) within the recording array.
    * @param position The position to set.
    */
-   inline void setPosition(const std::array<float, 3>& position) { m_position = position; }
+  inline void setPosition(const std::array<float, 3>& position)
+  {
+    m_position = position;
+  }
 
-   /**
+  /**
    * @brief Set index of channel across the recording system.
    * @param globalIndex The globalIndex to set.
    */
-   inline void setGlobalIndex(const SizeType globalIndex) { m_globalIndex = globalIndex; }
+  inline void setGlobalIndex(const SizeType globalIndex)
+  {
+    m_globalIndex = globalIndex;
+  }
 
-   /**
+  /**
    * @brief Set index of channel within the recording array.
    * @param localIndex The localIndex to set.
    */
-   inline void setLocalIndex(const SizeType localIndex) { m_localIndex = localIndex; }
+  inline void setLocalIndex(const SizeType localIndex)
+  {
+    m_localIndex = localIndex;
+  }
 
-   /**
+  /**
    * @brief Set index of array group the channel belongs to.
    * @param groupIndex The groupIndex to set.
    */
-   inline void setGroupIndex(const SizeType groupIndex) { m_groupIndex = groupIndex; }
+  inline void setGroupIndex(const SizeType groupIndex)
+  {
+    m_groupIndex = groupIndex;
+  }
 
-   /**
+  /**
    * @brief Set name of the channel.
    * @param name The name to set.
    */
-   inline void setName(const std::string& name) { m_name = name; }
+  inline void setName(const std::string& name) { m_name = name; }
 
-   /**
+  /**
    * @brief Set name of the array group the channel belongs to.
    * @param groupName The groupName to set.
    */
-   inline void setGroupName(const std::string& groupName) { m_groupName = groupName; }
+  inline void setGroupName(const std::string& groupName)
+  {
+    m_groupName = groupName;
+  }
 
-   /**
+  /**
    * @brief Set conversion factor.
    * @param conversion The conversion to set.
    */
-   inline void setConversion(const float conversion) { m_conversion = conversion; }
+  inline void setConversion(const float conversion)
+  {
+    m_conversion = conversion;
+  }
 
-   /**
+  /**
    * @brief Set sampling rate of the channel.
    * @param samplingRate The samplingRate to set.
    */
-   inline void setSamplingRate(const float samplingRate) { m_samplingRate = samplingRate; }
+  inline void setSamplingRate(const float samplingRate)
+  {
+    m_samplingRate = samplingRate;
+  }
 
-   /**
+  /**
    * @brief Set floating point value of microvolts per bit
    * @param bitVolts The bitVolts to set.
    */
-   inline void setBitVolts(const float bitVolts) { m_bitVolts = bitVolts; }
+  inline void setBitVolts(const float bitVolts) { m_bitVolts = bitVolts; }
 
 private:
   /**
@@ -182,12 +206,12 @@ private:
    */
   std::string m_comments;
 
-   /**
+  /**
    * @brief Coordinates of channel (x, y, z) within the recording array.
    */
   std::array<float, 3> m_position;
 
-   /**
+  /**
    * @brief Index of channel across the recording system.
    */
   SizeType m_globalIndex;
--- a/src/Channel.hpp
+++ b/src/Channel.hpp
@@ -134,7 +134,6 @@ public:
     m_position = position;
   }
 
-
   /**
    * @brief Set name of the channel.
    * @param name The name to set.
--- a/src/Channel.hpp
+++ b/src/Channel.hpp
@@ -37,72 +37,194 @@ public:
   ~Channel();
 
   /**
+   * @brief Copy constructor
+   */
+   Channel(const Channel& other) = default;
+
+   /**
+    * @brief Move constructor
+    */
+   Channel(Channel&& other) = default;
+
+   /**
+    * @brief Assignment operator
+    */
+   Channel& operator=(const Channel& other) = default;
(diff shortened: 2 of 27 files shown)


oruebel:
@stephprince  This is a first pass at fixing the naming conventions for member variables to avoid shadowing of names. For the main NWB data types, I propose that I fix those on the read PR #85 since I'll need to modify those members for read anyways and modifying them here as well will only lead to tons of merge conflicts.

stephprince:
Sounds good. As I'm looking through I had a question. I thought `m_` was supposed to replace `this->` but in some places both are being used. I wasn't sure if there is a reason to use both of them?

src/BaseIO.cpp
```diff
@@ -31,22 +31,28 @@ const BaseDataType BaseDataType::DSTR = BaseDataType(T_STR, DEFAULT_STR_SIZE);
...
 
 bool BaseIO::isOpen() const
 {
-  return opened;
+  return m_opened;
+}
+
+std::string BaseIO::getFileName()
+{
+  return this->m_filename;
```
stephprince, line 50:
```suggestion
  return m_filename;
```

stephprince: Some getters are defined as inline functions in the `.hpp` file and some are defined in the `.cpp` file. Wondering if we want to choose one approach and make it consistent.

src/Channel.hpp
```diff
@@ -36,73 +36,219 @@ class Channel
...
+  }
+
+  /**
+   * @brief Set index of array group the channel belongs to.
+   * @param groupIndex The groupIndex to set.
+   */
+  inline void setGroupIndex(const SizeType groupIndex)
+  {
+    m_groupIndex = groupIndex;
+  }
```
stephprince, line 162: I'm not sure if we want the user to be able to modify the indexes or the group information for the channel after the object is instantiated.

src/hdf5/HDF5IO.cpp
```diff
@@ -31,7 +28,7 @@ HDF5IO::~HDF5IO()
 
 std::string HDF5IO::getFileName()
 {
-  return filename;
+  return this->m_filename;
```
stephprince, line 31: several cases in this file of using both `this->` and `m_` that I think we could simplify

src/nwb/RecordingContainers.cpp
```diff
@@ -14,15 +14,15 @@ RecordingContainers::~RecordingContainers() {}
 
 void RecordingContainers::addContainer(std::unique_ptr<Container> container)
 {
-  this->containers.push_back(std::move(container));
+  this->m_containers.push_back(std::move(container));
```
stephprince, line 17:
```suggestion
  m_containers.push_back(std::move(container));
```

src/nwb/RecordingContainers.cpp
```diff
@@ -14,15 +14,15 @@ RecordingContainers::~RecordingContainers() {}
...
 }
 
 Container* RecordingContainers::getContainer(const SizeType& containerInd)
 {
-  if (containerInd >= this->containers.size()) {
+  if (containerInd >= this->m_containers.size()) {
```
stephprince, line 22:
```suggestion
  if (containerInd >= m_containers.size()) {
```

src/nwb/RecordingContainers.cpp
```diff
@@ -14,15 +14,15 @@ RecordingContainers::~RecordingContainers() {}
...
     return nullptr;
   } else {
-    return this->containers[containerInd].get();
+    return this->m_containers[containerInd].get();
```
stephprince, line 25:
```suggestion
    return m_containers[containerInd].get();
```

src/nwb/RecordingContainers.cpp
```diff
@@ -79,3 +79,8 @@ Status RecordingContainers::writeSpikeEventData(const SizeType& containerInd,
 
   ses->writeSpike(numSamples, numChannels, data, timestamps);
 }
+
+SizeType RecordingContainers::size()
+{
+  return this->m_containers.size();
```
stephprince, line 85:
```suggestion
  return m_containers.size();
```

src/nwb/ecephys/ElectricalSeries.cpp
```diff
@@ -45,36 +45,39 @@ void ElectricalSeries::initialize()
...
   }
   samplesRecorded = SizeArray(channelVector.size(), 0);
 
   // make channel conversion dataset
   channelConversion = std::unique_ptr<BaseRecordingData>(
-      io->createArrayDataSet(BaseDataType::F32,
-                             SizeArray {1},
-                             chunkSize,
-                             getPath() + "/channel_conversion"));
+      this->m_io->createArrayDataSet(BaseDataType::F32,
```
stephprince, line 55:
```suggestion
      m_io->createArrayDataSet(BaseDataType::F32,
```

src/nwb/ecephys/ElectricalSeries.cpp
```diff
@@ -45,36 +45,39 @@ void ElectricalSeries::initialize()
...
+                                     getPath() + "/channel_conversion"));
   channelConversion->writeDataBlock(
       std::vector<SizeType>(1, channelVector.size()),
       BaseDataType::F32,
       &channelConversions[0]);
-  io->createCommonNWBAttributes(getPath() + "/channel_conversion",
-                                "hdmf-common",
-                                "",
-                                "Bit volts values for all channels");
+  this->m_io->createCommonNWBAttributes(getPath() + "/channel_conversion",
```
stephprince, line 63:
```suggestion
  m_io->createCommonNWBAttributes(getPath() + "/channel_conversion",
```

src/nwb/ecephys/ElectricalSeries.cpp
```diff
@@ -45,36 +45,39 @@ void ElectricalSeries::initialize()
...
+      this->m_io->createArrayDataSet(BaseDataType::I32,
+                                     SizeArray {1},
+                                     chunkSize,
+                                     getPath() + "/electrodes"));
   electrodesDataset->writeDataBlock(
       std::vector<SizeType>(1, channelVector.size()),
       BaseDataType::I32,
       &electrodeInds[0]);
-  io->createCommonNWBAttributes(
+  this->m_io->createCommonNWBAttributes(
```
stephprince, line 78:
```suggestion
  m_io->createCommonNWBAttributes(
```

src/nwb/ecephys/ElectricalSeries.cpp
```diff
@@ -45,36 +45,39 @@ void ElectricalSeries::initialize()
...
       getPath() + "/electrodes", "hdmf-common", "DynamicTableRegion", "");
-  io->createReferenceAttribute(
+  this->m_io->createReferenceAttribute(
```
stephprince, line 80:
```suggestion
  m_io->createReferenceAttribute(
```

src/nwb/file/ElectrodeGroup.cpp
```diff
@@ -24,9 +24,11 @@ void ElectrodeGroup::initialize()
...
   Container::initialize();
 
-  io->createCommonNWBAttributes(path, "core", "ElectrodeGroup", description);
-  io->createAttribute(location, path, "location");
-  io->createLink("/" + path + "/device", "/" + device.getPath());
+  this->m_io->createCommonNWBAttributes(
+      this->m_path, "core", "ElectrodeGroup", description);
+  this->m_io->createAttribute(location, this->m_path, "location");
+  this->m_io->createLink("/" + this->m_path + "/device",
+                         "/" + device.getPath());
```
stephprince, line 31:
```suggestion
  m_io->createCommonNWBAttributes(
      m_path, "core", "ElectrodeGroup", description);
  m_io->createAttribute(location, m_path, "location");
  m_io->createLink("/" + m_path + "/device",
                         "/" + device.getPath());
```

src/nwb/file/ElectrodeTable.cpp
```diff
@@ -24,26 +24,30 @@ void ElectrodeTable::initialize()
...
+  groupNamesDataset->setDataset(std::unique_ptr<BaseRecordingData>(
+      this->m_io->createArrayDataSet(BaseDataType::STR(250),
+                                     SizeArray {0},
+                                     SizeArray {1},
+                                     this->m_path + "group_name")));
+  locationsDataset->setDataset(std::unique_ptr<BaseRecordingData>(
+      this->m_io->createArrayDataSet(BaseDataType::STR(250),
+                                     SizeArray {0},
+                                     SizeArray {1},
+                                     this->m_path + "location")));
```
stephprince, line 41:
```suggestion
  electrodeDataset->setDataset(std::unique_ptr<BaseRecordingData>(
      m_io->createArrayDataSet(BaseDataType::I32,
                                     SizeArray {1},
                                     SizeArray {1},
                                     m_path + "id")));
  groupNamesDataset->setDataset(std::unique_ptr<BaseRecordingData>(
      m_io->createArrayDataSet(BaseDataType::STR(250),
                                     SizeArray {0},
                                     SizeArray {1},
                                     m_path + "group_name")));
  locationsDataset->setDataset(std::unique_ptr<BaseRecordingData>(
      m_io->createArrayDataSet(BaseDataType::STR(250),
                                     SizeArray {0},
                                     SizeArray {1},
                                     m_path + "location")));
```

src/nwb/NWBFile.cpp
```diff
@@ -41,44 +42,44 @@ NWBFile::~NWBFile() {}
...
-  io->createGroup("/acquisition");
-  io->createGroup("/analysis");
-  io->createGroup("/processing");
-  io->createGroup("/stimulus");
-  io->createGroup("/stimulus/presentation");
-  io->createGroup("/stimulus/templates");
-  io->createGroup("/general");
-  io->createGroup("/general/devices");
-  io->createGroup("/general/extracellular_ephys");
+  this->m_io->createCommonNWBAttributes("/", "core", "NWBFile", "");
```
stephprince, line 65: several cases using both `this->` and `m_io` in this file that I think could be simplified to `m_io`

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -21,9 +21,9 @@ void DynamicTable::initialize()
 {
   Container::initialize();
 
-  io->createCommonNWBAttributes(
-      path, "hdmf-common", "DynamicTable", getDescription());
-  io->createAttribute(getColNames(), path, "colnames");
+  this->m_io->createCommonNWBAttributes(
+      this->m_path, "hdmf-common", "DynamicTable", getDescription());
+  this->m_io->createAttribute(getColNames(), this->m_path, "colnames");
```
stephprince, line 26:
```suggestion
  m_io->createCommonNWBAttributes(
      m_path, "hdmf-common", "DynamicTable", getDescription());
  m_io->createAttribute(getColNames(), m_path, "colnames");
```

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -32,30 +32,30 @@ void DynamicTable::addColumn(const std::string& name,
...
     for (SizeType i = 0; i < values.size(); i++)
-      vectorData->dataset->writeDataBlock(
+      vectorData->m_dataset->writeDataBlock(
           std::vector<SizeType>(1, 1),
           BaseDataType::STR(values[i].size() + 1),
           values[i].c_str());  // TODO - add tests for this
-    io->createCommonNWBAttributes(
-        path + name, "hdmf-common", "VectorData", colDescription);
+    this->m_io->createCommonNWBAttributes(
+        this->m_path + name, "hdmf-common", "VectorData", colDescription);
```
stephprince, line 45:
```suggestion
    m_io->createCommonNWBAttributes(
        m_path + name, "hdmf-common", "VectorData", colDescription);
```

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -32,30 +32,30 @@ void DynamicTable::addColumn(const std::string& name,
...
+  if (!elementIDs->isInitialized()) {
     std::cerr << "ElementIdentifiers dataset is not initialized" << std::endl;
   } else {
-    elementIDs->dataset->writeDataBlock(
+    elementIDs->m_dataset->writeDataBlock(
         std::vector<SizeType>(1, values.size()), BaseDataType::I32, &values[0]);
-    io->createCommonNWBAttributes(
-        path + "id", "hdmf-common", "ElementIdentifiers");
+    this->m_io->createCommonNWBAttributes(
+        this->m_path + "id", "hdmf-common", "ElementIdentifiers");
```
stephprince, line 58:
```suggestion
    m_io->createCommonNWBAttributes(
        m_path + "id", "hdmf-common", "ElementIdentifiers");
```

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -66,9 +66,9 @@ void DynamicTable::addColumn(const std::string& name,
   if (values.empty()) {
     std::cerr << "Data to add to column is empty" << std::endl;
   } else {
-    io->createReferenceDataSet(path + name, values);
-    io->createCommonNWBAttributes(
-        path + name, "hdmf-common", "VectorData", colDescription);
+    this->m_io->createReferenceDataSet(this->m_path + name, values);
+    this->m_io->createCommonNWBAttributes(
+        this->m_path + name, "hdmf-common", "VectorData", colDescription);
```
stephprince, line 71:
```suggestion
    m_io->createReferenceDataSet(m_path + name, values);
    m_io->createCommonNWBAttributes(
        m_path + name, "hdmf-common", "VectorData", colDescription);
```

stephprince (commented):
Added a couple comments/questions! Mainly suggestions to use `m_` instead of both `this->` and `m_` since I think they are redundant? If there is a reason to use both feel free to ignore those.

oruebel:
> Mainly suggestions to use `m_` instead of both `this->` and `m_` since I think they are redundant? If there is a reason to use both feel free to ignore those.

In this case, yes, `this->m_` and `m_` access the same variable. `this` is a pointer to the current object of the class so it makes it clear that we are using the object and `m_` makes it clear that it is a class member variable. We can just use `m_` in most places if you prefer.

src/BaseIO.cpp
oruebel: For simple getter functions,I think it is generally useful to define them as inline methods in the header file as it: 1) reduces overhead and improves performance, because the compiler usually can expand inline functions directly at compile time so you don't have the overhead of the function call, and 2) it helps keep the code readable and avoid redundant text in the header and cpp file. For more complex getter functions that involve validation, performing data retrieval, or other complex operations, I think it is best to implement those in the cpp file, to keep the header file easy to read and also expanding complex functions at compile can become expansive and is not as effective.

src/Channel.hpp
oruebel: Sounds reasonable. Please let me know which setters you would like me to remove or feel free to just make edits to the PR directly.

src/hdf5/HDF5IO.cpp
oruebel: I've now replaced all occurrences of `this->m_` with just `m_`

src/Channel.hpp
stephprince, line 162:
```suggestion
```

src/Channel.hpp
```diff
@@ -36,73 +36,219 @@ class Channel
...
+  inline void setSamplingRate(const float samplingRate)
+  {
+    m_samplingRate = samplingRate;
+  }
+
+  /**
+   * @brief Set floating point value of microvolts per bit
+   * @param bitVolts The bitVolts to set.
+   */
+  inline void setBitVolts(const float bitVolts) { m_bitVolts = bitVolts; }
```
stephprince, line 201:
```suggestion
  /**
   * @brief Set name of the channel.
   * @param name The name to set.
   */
  inline void setName(const std::string& name) { m_name = name; }
```

src/Channel.hpp
stephprince: Just pushed a change with the setters I think we should remove for now.

stephprince (approved):

oruebel:
@stephprince can you please re-review. I moved some of the simple setter functions.

stephprince (approved):
Looks good to me!

oruebel:
@stephprince is it expected that the validate tests are failing or is this new in this PR. Just wanted to check if this is something I need to fix here or whether this is something for a separate PR.

oruebel:
Ok, comparing the test files generated by the main branch and this branch it looks like something broke with writing the `colnames` for `DynamicTable` and/or `ElectrodeTable`

oruebel:
@stephprince can you re-review. All tests are passing now. 

Turns out, the issue with the invalid `colnames` in the created NWB files was due to https://github.com/NeurodataWithoutBorders/aqnwb/issues/106 . I had missed that ``ElectrodeTable`` defined its own `ElectordeTable.colNames` which was hiding the inherited member `DynamicTable.colNames` and because I had only renamed one of them, it was using the empty `colNames` from `DynamicTable` instead of from `ElectrodeTable`.

stephprince:
ah ok. So from what I understand, #106 is resolved but is a behavior to keep an eye out for in the future

stephprince (approved):

oruebel:
> So from what I understand, #106 is resolved but is a behavior to keep an eye out for in the future

Correct. At least I hope I caught all the places where this is happening.

Issue #96, opened 2024-09-11
Fix validation workflow to throw errors on invalid files


stephprince:
The `validate` step in the `test.yml` workflow currently only print the nwbinspector report summary and will pass even if the inspector emits critical error warnings. 

We should decide on a level of importance to use as a threshold and have the workflow fail when test files have issues.

PR #118, opened 2024-12-26
Add HDF5IO tests for add_read PR

oruebel:
Add unit tests for public ``HDF5IO``-specifc functions (i.e., not inherited from ``BaseIO``. These may be moved to a separate PR:
    - [X] ``HDF5IO::getH5ObjectType``
    - [ ] ``HDF5IO::getNativeType``
    - [ ] ``HDF5IO::getH5Type``

### DIFF
--- a/src/io/hdf5/HDF5IO.hpp
+++ b/src/io/hdf5/HDF5IO.hpp
@@ -311,7 +311,7 @@ public:
   /**
    * @brief Returns the HDF5 type of object at a given path.
    * @param path The location in the file of the object.
-   * @return The type of object at the given path.
+   * @return The type of object at the given path. A negative value indicates that an error occurred.
    */
   H5O_type_t getH5ObjectType(const std::string& path) const;
 
--- a/tests/testHDF5IO.cpp
+++ b/tests/testHDF5IO.cpp
@@ -509,3 +509,33 @@ TEST_CASE("readDataset", "[hdf5io]")
     hdf5io->close();
   }
 }
+
+TEST_CASE("getH5ObjectType", "[hdf5io]")
+{
+  // create and open file
+  std::string filename = getTestFilePath("test_getH5ObjectType.h5");
+  IO::HDF5::HDF5IO hdf5io(filename);
+  hdf5io.open();
+
+  SECTION("group")
+  {
+    hdf5io.createGroup("/group");
+    REQUIRE(hdf5io.getH5ObjectType("/group") == H5O_TYPE_GROUP);
+  }
+
+  SECTION("dataset")
+  {
+    std::vector<int> testData = {1, 2, 3, 4, 5};
+    std::string dataPath = "/dataset";
+    hdf5io.createArrayDataSet(BaseDataType::I32, SizeArray {0}, SizeArray {1}, dataPath);
+    REQUIRE(hdf5io.getH5ObjectType(dataPath) == H5O_TYPE_DATASET);
+  }
+
+  SECTION("non_existent_object")
+  {
+    REQUIRE(hdf5io.getH5ObjectType("/non_existent") < 0); 
+  }
+
+  // close file
+  hdf5io.close();
+}


oruebel:
@stephprince there is still an issue with reading strings, but otherwise this PR is ready for you to take a look at.

oruebel:
I have confirmed now that the fixed length string dataset is being written correctly. However, reading of fixed length string data still fails. I have not investigated the issues with variable length strings yet.

oruebel:
Ok, fixed length strings seem to work now for both write and read. Next up, variable length strings.

oruebel:
@stephprince I've now fixed the writing/reading of fixed and variable length strings. This PR is ready for you to take a look at. I think the other TODO items for #85 are probably best done in additional PRs, otherwise this PR will become to large. 

The last remaining issue for this PR is to fix the address sanitizer test. It looks like the sanitizer doesn't like something about the changes in ``HDF5RecordingData::writeDataBlock`` . Getting string datasets to work for write/read is a bit tricky.

src/io/hdf5/HDF5RecordingData.cpp
```diff
@@ -60,51 +60,101 @@ Status HDF5RecordingData::writeDataBlock(
...
-    m_dataset->write(data, nativeType, mSpace, fSpace);
+    DataType nativeType;
+    if (type.type == BaseDataType::Type::V_STR) {
+      // Create a variable-length string type
+      nativeType = StrType(0, H5T_VARIABLE);
+    } else {
+      nativeType = HDF5IO::getNativeType(type);
+    }
+
+    if (type.type == BaseDataType::Type::V_STR) {
```
stephprince, line 99: Just to clarify, is this functionality needed in addition to the ``Status HDF5IO::createStringDataSet(const std::string& path, const std::vector<std::string>& values)`` function or intended to replace it?

stephprince (commented):
looks good to me! thanks for adding all the tests

src/io/hdf5/HDF5RecordingData.cpp
oruebel: Good point.  I think the key difference is that `createStringDataSet` is a one-shot write, i.e., it creates and writes the data in the same call. Using `HDF5RecoringData` however writes data in blocks, so you can create an array of strings and write them as you go. As such, `createStringDataSet` is nice for creating metadata datasets where the values are known, and `HDF5RecordingData::writeStringDataBlock` (combined with `HDF5IO::createArrayDataSet`) is useful when acquiring string data. 

TBH, I was creating unit tests for all main data types to test reading and writing and simply tried to follow the same pattern for all data types, and didn't realize that string data was following a different pattern. That being said, I think it is useful to be able to acquire string data in blocks as well. 

I guess the question is, whether we want to: a) always go the route of first calling create an array dataset with ``HDF5IO::createArrayDataset`` and then writing the data with ``BaseRecordingData::writeDataBlock`` calls or b) have additional functions on HDF5IO for one-shot writes where we create and write the dataset in the  same call. I think in principle option b is fine, with the caveat that we have one-shot write functions only for a few special types like strings and references. 

What do you think?

oruebel:
> The last remaining issue for this PR is to fix the address sanitizer test. It looks like the sanitizer doesn't like something about the changes in `HDF5RecordingData::writeDataBlock` . Getting string datasets to work for write/read is a bit tricky.

To get writing of string data to work cleanly, I added a new function `HDF5RecordingData::writeStringDataBlock` to avoid the error-prone casting of `void*` back to ``std::vector<std::string>``. This has fixed the address sanitizer issues, so this PR is now good for final review. 

@stephprince the change of adding `HDF5RecordingData::writeStringDataBlock`  also relates directly to the question you raised above, i.e., do we want to support acquiring string data in blocks or is it ok to only support one-shot writing of strings via `HDF5IO::createStringDataSet`?

src/io/hdf5/HDF5RecordingData.cpp
stephprince: > As such, createStringDataSet is nice for creating metadata datasets where the values are known, and HDF5RecordingData::writeStringDataBlock (combined with HDF5IO::createArrayDataSet) is useful when acquiring string data.

Thanks for clarifying! That makes sense.

> I guess the question is, whether we want to: a) always go the route of first calling create an array dataset with HDF5IO::createArrayDataset and then writing the data with BaseRecordingData::writeDataBlock calls or b) have additional functions on HDF5IO for one-shot writes where we create and write the dataset in the same call

I also think option B is fine and it is useful to have that functionality for small datasets that are known before the recording starts. I agree we should probably limit it to a few special types and in most cases follow the 1) create an array dataset with `HDF5IO::createArrayDataset` and 2) write the data with `BaseRecordingData::writeDataBlock` pattern. 

If we go with option B, we may want to update `createStringDataset` to use `writeStringDataBlock` to make it consistent and easier to update string dataset writing in the future. Similarly for `createReferenceDataset`, though I think we do not have testing for writing datasets of references yet(?) so I'm not sure if that's fully supported or would also require a `writeReferenceDataBlock` method.

oruebel: > I also think option B is fine

This sounds all good to me.

> we may want to update `createStringDataset` to use `writeStringDataBlock` to make it consistent and easier to update string dataset writing in the future.

Makes sense

> Similarly for `createReferenceDataset`, though I think we do not have testing for writing datasets of references yet(?) so I'm not sure if that's fully supported or would also require a `writeReferenceDataBlock` method.

I think you are correct that `HDF5RecordingData::writeDataBlock` probably doesn't support writing references yet and that we may need to handle that as a a separate `HDF5RecordingData::writeReferenceDataBlock`. I think we can make that a separate issue though.

stephprince: yes, definitely think we can handle writing references with `writeDataBlock` in a separate PR.

stephprince:
I was reminded as I was working on a different PR that I still need to fix the nwbinspector validate workflow because it will currently pass even if there are critical errors (see https://github.com/NeurodataWithoutBorders/aqnwb/issues/96). 

I'll fix the workflow, but in the meantime just wanted to note that there seem to be some validation issues introduced in this PR (and the base read PR) even though the tests are passing.

oruebel:
How do I need to run the NWB Inspector locally to see the validation errors?

stephprince:
Run `nwbinspector . --threshold BEST_PRACTICE_VIOLATION` in the build folder containing the test data (e.g. build/dev/tests/data). It should inspect only the files ending in `.nwb`

PR #122, opened 2025-01-07
Update nwbinspector validation workflow

### LINKED ISSUE #96: Fix validation workflow to throw errors on invalid files

stephprince:
Fix #96. 

The validation job will now fail if NWBInspector finds issues at or above the BEST_PRACTICE_VIOLATION importance level in the NWB files generated by the tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -141,8 +141,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install nwbinspector
-        nwbinspector nwb_files --threshold BEST_PRACTICE_VIOLATION --json-file-path out.txt
-        if ! grep -q '"messages": \[\]' out.txt; then
+        nwbinspector nwb_files --threshold BEST_PRACTICE_VIOLATION --json-file-path out.json
+        if ! grep -q '"messages": \[\]' out.json; then
           echo "NWBInspector found issues in the NWB files"
           exit 1
         fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -141,4 +141,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install nwbinspector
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -141,4 +141,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install nwbinspector
-        nwbinspector nwb_files --threshold BEST_PRACTICE_VIOLATION
+        nwbinspector nwb_files --threshold BEST_PRACTICE_VIOLATION --json-file-path out.txt
+        if ! grep -q '"messages": \[\]' out.txt; then
+          echo "NWBInspector found issues in the NWB files"
```
oruebel, line 146: It would be nice to also print the contents of the out.txt file to the command-line to make it easy to see the issues found in the GitHub Action directly. I would also rename out.txt to out.json

oruebel (changes requested):

.github/workflows/tests.yml
oruebel, line 146:
```suggestion
        nwbinspector nwb_files --threshold BEST_PRACTICE_VIOLATION --json-file-path out.json
        if ! grep -q '"messages": \[\]' out.json; then
          echo "NWBInspector found issues in the NWB files"
          cat out.json
```

stephprince: > It would be nice to also print the contents of the out.txt file to the command-line to make it easy to see the issues found in the GitHub Action directly.

The report summary should be printed out and formatted by running nwbinspector, in this action there are 0 issues found but otherwise it will print the issues in the GitHub Action.

>  I would also rename out.txt to out.json

Will do!

oruebel (approved):

PR #123, opened 2025-01-08
Update read example

oruebel:
#118 fixed several issues with read so this PR updates the read example to resolve a few items that were marked as TODO in the example that are now working.


oruebel:
Changed the base branch to #128 temporarily to see if the changes there fix the issues here. We can change back to target #85 again later.

oruebel:
Ok, I finally figured out the Windows issue. The problem was that the `espath` used for reading was determined in this loop:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/14a825c4f054f468418cad7657f789e4ee1ccdd0/tests/examples/test_ecephys_data_read.cpp#L203-L208

On Windows the order of paths is different and so the wrong ElectricalSeries was used, and since the size of the two series was different this led to an out-of-bounds error.

oruebel:
@stephprince this PR is now also ready for review

oruebel:
@stephprince I merged with #85 just so that I can finalize things there

PR #124, opened 2025-01-09
Add read tests for `TimeSeries` , VectorData`, `Data`, `ElementIdentifers`, `Device`

### LINKED ISSUE #32: Save additional TimeSeries fields
Add additional fields to the TimeSeries class from the [schema](https://nwb-schema.readthedocs.io/en/latest/format.html#timeseries):
- [ ] continuity
- [x] offset
- [ ] control
- [ ] control_description
- [ ] sync(?)
- [ ] starting_time(?)

### LINKED ISSUE #126: Update handling of Data and VectorData
1. Move creation of neurodata_type and namespace attributes to Data and add DEFINE_FIELD for those (similar to Container)
2. Add initalize() methods for ``Data`` and ``VectorData``
3. Update ``DynamicTable`` to call ``VectorData.initalize`` instead of creating the `VectorData` manually

oruebel:
Add unit tests for readfor `TimeSeries` For every neurodata_type we should test that: a) the type itself can be read from file, and b) that all ``DEFINE_FIELDS`` specified for read can be read. 
- [X] Add base read test for `TimeSeries`

Fix #32 As part of writing the tests for TimeSeries, we should also double-check that all fields that AqNWB writes, that they are also exposed for read and that all types listed for read are also written.
- [X] Add missing `TimeSeries.data.offset` attribute on write
- [X] Add missing `TimeSeries.data.continuity` attribute on write
- [X] Add missing fields `control` and `control_description` datasets on write (read via ``readControlDescription`` and``readControl``)
- [X] Add missing field `starting_time` dataset and corresponding attributes `unit` and `rate` on write (read via  `readStartingTime`, `readStartingTimeUnit`, `readStartingTimeRate`
- [X] Move the creation of neurodata_type and namespace attributes to Container and add corresponding DEFINE_FIELDs for read

Fix #126 handling of ``Data`` and ``VectorData``:
- [X]  Moved creation of neurodata_type and namespace attributes to Data and added DEFINE_FIELD for those (similar to Container)
- [X] Added ``initalize()`` methods for ``Data`` and ``VectorData``
- [X] Update ``DynamicTable`` and ``ElectrodeTable`` to call ``VectorData.initalize`` instead of setting up ``VectorData`` manually

Added write/read tests for additional types
- [X] `VectorData` with `int`, `double`, and `string` data
- [X] `Data` with a single type (the remaining is covered by the `VectorData` tests)
- [X] `ElementIdentifiers` and updated the class to overwrite the `readData` method to set the data type to int by default
- [X] `Device`

Additional fixes needed:
- [x] Fix handling on `controlDescription` see https://github.com/NeurodataWithoutBorders/aqnwb/pull/124#discussion_r1911223968


oruebel:
@stephprince this PR is now ready for you to take a look at

src/nwb/base/TimeSeries.cpp
```diff
@@ -25,54 +74,124 @@ void TimeSeries::initialize(const IO::BaseDataType& dataType,
...
+        positionOffset[0]};
+    IO::BaseDataType controlDesriptionType(IO::BaseDataType::Type::V_STR,
+                                           0);  // 0 indicates variable length
+    tsStatus = this->control_description->writeStringDataBlock(
+        controlDescriptionShape,
+        controlDescriptionPositionOffset,
+        controlDesriptionType,
+        controlDescriptionInput);
+  }
+
```
stephprince, line 194: From my understanding of the schema, controlDescription should not be the shape of the first data dimension, but should instead be the number of unique control values provided. 

It is also a little complicated where I think controlDescription must be provided if control is present. However you would only need to write the controlDescription dataset once and not necessarily each time you are writing to the control dataset, so I don't think the TimeSeries::writeData is where we would necessarily want to place those checks.

src/nwb/base/TimeSeries.hpp
```diff
@@ -41,10 +72,13 @@ class TimeSeries : public Container
...
-                   const std::vector<SizeType>& positionOffset,
-                   const void* dataInput,
-                   const void* timestampsInput = nullptr);
+  Status writeData(
+      const std::vector<SizeType>& dataShape,
+      const std::vector<SizeType>& positionOffset,
+      const void* dataInput,
+      const void* timestampsInput = nullptr,
+      const void* controlInput = nullptr,
+      const std::vector<std::string>& controlDescriptionInput = {});
```
stephprince, line 81: I think `RecordingContainers::writeTimeSeriesData` should also be updated to accept and pass along the `controlInput` argument

src/io/hdf5/HDF5IO.cpp
```diff
@@ -1177,3 +1224,46 @@ H5::DataType HDF5IO::getH5Type(IO::BaseDataType type)
...
+        // Array type: get the base type and size
+        H5::ArrayType arrayType(nativeType);
+        H5::DataType baseType = arrayType.getSuper();
+        hsize_t size = arrayType.getArrayNDims();
+        return IO::BaseDataType(getBaseType(baseType).type, size);
+    } else {
+        // Default case: return a 32-bit integer type
+        return IO::BaseDataType(IO::BaseDataType::Type::T_I32);
+    }
+}*/
```
stephprince, line 1269: I think this is a duplicate of `HDF5IO::getBaseDataType` and can be removed

```suggestion

```

tests/testTimeSeries.cpp
```diff
@@ -0,0 +1,328 @@
...
+  BaseDataType dataType = BaseDataType::F32;
+  std::vector<float> data = getMockData1D(numSamples);
+  BaseDataType timestampsType = BaseDataType::F64;
+  std::vector<double> timestamps = getMockTimestamps(numSamples, 1);
+  std::vector<unsigned char> controlData = {0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
+  std::vector<std::string> controlDescription = {
+      "c0", "c1", "c0", "c1", "c0", "c1", "c0", "c1", "c0", "c1"};
+  std::string path = getTestFilePath("testTimeseries.h5");
+
+  SECTION("test writing timeseries data block")
```
stephprince, line 30:
```suggestion
  SECTION("test reading timeseries data block and attributes")
```

tests/testTimeSeries.cpp
```diff
@@ -0,0 +1,328 @@
...
+        AQNWB::NWB::RegisteredType::create(fullClassName, dataPath, io);
+    std::string containerType = readContainer->getTypeName();
+    REQUIRE(containerType == "TimeSeries");
+
+    // Open the TimeSeries container directly from file using the utility method
+    // This method does the same steps as above, i.e., read the attributes and
+    // then create the type from the given name
+    auto readTS = AQNWB::NWB::RegisteredType::create(dataPath, io);
+    std::string readTSType = readContainer->getTypeName();
+    REQUIRE(readTSType == "TimeSeries");
```
stephprince, line 91: I might be missing a distinction here, but I think these tests are duplicated in the section below and these examples could maybe be better included as part of the registeredType tests

stephprince (commented):
I've added my review comments for the TimeSeries tests, but I have not gotten to the Vector Data or Table related changes yet

src/nwb/base/TimeSeries.cpp
oruebel: Yes, you are correct. I misread the schema. Ok, we'll need to fix that.

tests/testTimeSeries.cpp
oruebel: I updated and moved the test to testRegisteredType

stephprince:
Finished looking through the other files, other than the previous comments all looks good to me!

src/nwb/base/TimeSeries.cpp
oruebel: I updated the code accordingly. 
* The initial `control_description` values are now passed directly to `TimeSeries.initialize` 
* `TimeSeries.writeData` no longer takes `control_description` as input (but still accepts `control`)
* Users can update the `TimeSeries.control_description` manually if they need to modify values after `TimeSeries.initialize`

src/nwb/base/TimeSeries.hpp
oruebel: I believe `RecordingContainers::writeElectricalSeriesData` and `RecordingContainers::writeSpikeEventData` also need the same update.

oruebel: Done

PR #125, opened 2025-01-09
address validation errors in add_read

stephprince:
I looked more closely at the nwbinspector errors and most were due to the updates to the string dataset write.

I believe the remaining critical error is actually an issue on the nwbinspector end. According to the [schema](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/d5a931d2cc0ae059d917705b7583136d756a2c8b/core/nwb.ecephys.yaml#L68-L84), SpikeEventSeries can have dimensions [nevents, nsamples] when it contains data from a single channel, but the inspector will throw an error that the second dimension does not match the number of electrodes: 

```
0  CRITICAL
===========

0.0  createESandSES.nwb: check_electrical_series_dims - 'SpikeEventSeries' object at location '/acquisition/spikedata1'
       Message: The second dimension of data does not match the length of electrodes. Your data may be transposed.
```

I can fix the issue here by changing the test to use multiple channels, but will also open a related issue on the nwbinspector side

### DIFF
--- a/src/io/hdf5/HDF5IO.cpp
+++ b/src/io/hdf5/HDF5IO.cpp
@@ -898,17 +898,14 @@ Status HDF5IO::createStringDataSet(const std::string& path,
   if (!m_opened)
     return Status::Failure;
 
-  std::vector<const char*> cStrs;
-  cStrs.reserve(values.size());
-  for (const auto& str : values) {
-    cStrs.push_back(str.c_str());
-  }
-
   std::unique_ptr<IO::BaseRecordingData> dataset;
   dataset = std::unique_ptr<IO::BaseRecordingData>(createArrayDataSet(
       IO::BaseDataType::V_STR, SizeArray {values.size()}, SizeArray {1}, path));
-  dataset->writeDataBlock(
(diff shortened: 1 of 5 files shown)


stephprince:
@oruebel this is ready for your review to merge into the add_read branch

One thing I was wondering was instead of having `writeDataBlock` and `writeStringDataBlock`, if we would want to overload `writeDataBlock` instead. We could keep the warning if a string datatype is provided in the wrong manner, but in this way we would provide a consistent interface whether the user is providing raw data or string data. I am fine keeping it as is but just wanted to suggest as another option.

oruebel:
> One thing I was wondering was instead of having `writeDataBlock` and `writeStringDataBlock`, if we would want to overload `writeDataBlock` instead. We could keep the warning if a string datatype is provided in the wrong manner, but in this way we would provide a consistent interface whether the user is providing raw data or string data. I am fine keeping it as is but just wanted to suggest as another option.

Overloading `writeDataBlock` instead is fine. I was just having a lot of trouble with getting a void pointer back to a proper vector of strings without running into address sanitizer issues. I created issue #127 for this so we can address it separately.

oruebel (approved):

PR #128, opened 2025-01-12
Make calling io->open() explicit

oruebel:
While working on debugging #123 I came across the issue that `open` seemed to be called more than once on the same file. Part of the issue seems to be that `io->open` is being called implicitly on behalf of the user in `NWBFile.initialize`. This PR updates the logic to always require that the user call `io-open` before making any modifications to the file, e.g, initializing the `NWBFile`. This PR makes a few changes:

- [X] Modified `NWBFile.initialize` to more robustly check if a file has been initialized. The current logic requires that a file must not exist on disk before calling `NWBFile.initialize`, i.e., if a user calls `io->open()` before calling `NWBFile.initialize`, the current logic would fail.
- [X] Added `NWBFile.isInitialized` to check whether the expected file structure exists within the file
- [X] Updated `HDF5IO.create*` methods to perform the more strict check that `canModifyObjects()` is true rather than just whether the file `isOpen()`
- [X] Updated test suite to call `io->open()` directly where it was missing
- [X] Updated `testWorkflowExamples.cpp` to add `REQUIRE` checks . The example previously passed independent of whether the workflow was successful or not because the status of I/O operations was not being checked. Since the AqNWB functions did not raise exceptions (but simply returned a ``Status::Failure``) the workflow would run to completion even if the file was not being created at all.

### DIFF
--- a/src/io/hdf5/HDF5IO.cpp
+++ b/src/io/hdf5/HDF5IO.cpp
@@ -626,8 +626,9 @@ Status HDF5IO::createAttribute(const IO::BaseDataType& type,
   DataType H5type;
   DataType origType;
 
-  if (!m_opened)
+  if (!canModifyObjects()) {
     return Status::Failure;
+  }
 
   // open the group or dataset
   H5O_type_t objectType = getH5ObjectType(path);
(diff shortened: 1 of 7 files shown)


oruebel:

stephprince (approved):
These changes look good to me and I'm fine to make io->open() explicit. 

However, I'm not sure what the original issue was with calling io->open() twice? I thought it was handled in `Status HDF5IO::open(FileMode mode)` that if the file was already opened and io->open() was called, nothing would happen.

PR #135, opened 2025-01-13
Add codecov badge to README.md

oruebel:


stephprince (dismissed):
Looks good to me! However the overall code coverage percentage is still including the test suite currently. I'm not sure if we want to fix that first before including in the README to be more accurate

oruebel:
> Looks good to me! However the overall code coverage percentage is still including the test suite currently. I'm not sure if we want to fix that first before including in the README to be more accurate

Sounds good. Please just merge when this is ready .

stephprince:
Tested out some changes to the coverage command here, but was still running into issues on what folders were included in the report uploaded to CodeCov. Will address these in a separate issue.

stephprince (approved):

PR #137, opened 2025-01-14
Add unit tests for Utils.hpp

oruebel:


oruebel:
@stephprince note, I did not add unit tests for the ``convertFloatToInt16LE`` function because the behavior in edge cases was not clear to me and led to errors when I tried to create tests for it automatically with Cline.

stephprince (approved):

PR #138, opened 2025-01-14
Suppress boost warnings

### LINKED ISSUE #87: Suppress boost warnings in build
Currently we see a lot of warnings during compile generated by the boost library. To make it easier to focus on warnings and errors generated by AqNWB itself, it would be nice to suppress warnings generated by Boost. I think the``-isystem`` flag could work to tell the compiler to treat the BOOST directory as a system include directory, which should suppresses warnings from those headers. I.e., add something along the lines of ```COMPILE_OPTIONS "-isystem${Boost_INCLUDE_DIRS}"``` in the CMakeLists.txt

stephprince:
Fix #87 by treating boost directory as a system include directory.

Also fixes various warnings to clean up the build output.

### DIFF
--- a/src/io/hdf5/HDF5RecordingData.hpp
+++ b/src/io/hdf5/HDF5RecordingData.hpp
@@ -87,7 +87,6 @@ private:
    */
   Status writeDataBlockHelper(const std::vector<SizeType>& dataShape,
                               const std::vector<SizeType>& positionOffset,
-                              const AQNWB::IO::BaseDataType& type,
                               H5::DataSpace& mSpace,
                               H5::DataSpace& fSpace);
 
--- a/src/io/hdf5/HDF5RecordingData.hpp
+++ b/src/io/hdf5/HDF5RecordingData.hpp
@@ -80,7 +80,6 @@ private:
    * @brief Allocate space and validate parameters
    * @param dataShape The size of the data block.
    * @param positionOffset The position of the data block to write to.
-   * @param type The data type of the elements in the data block.
    * @param mSpace The HDF5 memory space (return value)
    * @param fSpace The HDF5 file space (return value)
    * @return The status of the write operation.
--- a/src/nwb/ecephys/ElectricalSeries.cpp
+++ b/src/nwb/ecephys/ElectricalSeries.cpp
@@ -26,6 +26,7 @@ void ElectricalSeries::initialize(const IO::BaseDataType& dataType,
                                   const std::string& description,
                                   const SizeArray& dsetSize,
                                   const SizeArray& chunkSize,
+                                  const float& conversion,
                                   const float& resolution,
                                   const float& offset)
 {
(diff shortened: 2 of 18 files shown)


src/io/hdf5/HDF5RecordingData.hpp
```diff
@@ -87,7 +87,6 @@ class HDF5RecordingData : public AQNWB::IO::BaseRecordingData
    */
   Status writeDataBlockHelper(const std::vector<SizeType>& dataShape,
                               const std::vector<SizeType>& positionOffset,
```
oruebel, line 89: Please also remove "* @param type The data type of the elements in the data block." from the docstring

src/nwb/ecephys/ElectricalSeries.cpp
```diff
@@ -26,7 +26,6 @@ void ElectricalSeries::initialize(const IO::BaseDataType& dataType,
                                   const std::string& description,
                                   const SizeArray& dsetSize,
                                   const SizeArray& chunkSize,
-                                  const float& conversion,
```
oruebel, line 29: Instead of removing conversion, I believe the error is that it should be passed to `TimeSeries.initialize` at the beginning of the function

src/nwb/ecephys/SpikeEventSeries.cpp
```diff
@@ -30,7 +29,6 @@ void SpikeEventSeries::initialize(const IO::BaseDataType& dataType,
                                description,
                                dsetSize,
                                chunkSize,
-                               conversion,
```
oruebel, line 33: Note, this will change the order of arguments. `conversion` appears before `resolution` and `offset` in `TimeSeries.initalize` so by not providing a conversion factor at all `resolution` will be assigned to `conversion` and `offset` to `resolution`. If we don't want to use the `conversion` attribute then we should add logic to set conversion to `NaN` and updated `TimeSeries` to ignore `conversion` if it is set to `NaN`.

oruebel (changes requested):
I added a few comments regarding the use of `conversion` in the different TimeSeries

src/nwb/ecephys/ElectricalSeries.cpp
stephprince: I see, the conversion value passed to `TimeSeries.initialize` was previously obtained from the `channelVector` data structure, so the `conversion` input argument was unused in both ElectricalSeries and SpikeEventSeries.

I didn't realize that [according to the schema](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/d5a931d2cc0ae059d917705b7583136d756a2c8b/core/nwb.ecephys.yaml#L40C18-L42C66), the overall conversion value should still be used even if there are individual channel conversions. So yes it should be passed to `TimeSeries.initialize` instead of the channelVector conversion value (similarly for SpikeEventSeries).

src/nwb/ecephys/SpikeEventSeries.cpp
stephprince: See the changes I made above. Conversion was previously provided to `TimeSeries.initialize` from the channelVector data structure in `ElectricalSeries.initialize`, but I think it should instead just be updated to use the conversion argument to `SpikeEventSeries.initialize` or `ElectricalSeries.initialize`

oruebel (approved):

Issue #97, opened 2024-09-11
add AnnotationSeries datatype


stephprince:
Add the AnnotationSeries datatype to hold text information from the acquisition system. 

Follow the general steps for adding a new data type:
- [ ] add a class for `AnnotationSeries`
- [ ] add a `createAnnotationSeries` function on `NWBFile`
- [ ] add a `writeAnnotationData` function on `recordingContainers`
- [ ] add a `writeText` or relevant convenience function on `AnnotationSeries` if needed

PR #141, opened 2025-01-15
Add AnnotationSeries data type

stephprince:
Fix https://github.com/NeurodataWithoutBorders/aqnwb/issues/97. These changes:
* add the AnnotationSeries class
* add convenience methods for creating an AnnotationSeries dataset and adding annotations to the file
* add testing for the class and convenience methods

### DIFF
--- a/src/nwb/misc/AnnotationSeries.hpp
+++ b/src/nwb/misc/AnnotationSeries.hpp
@@ -46,13 +46,13 @@ public:
   /**
    * @brief Writes a channel to an AnnotationSeries dataset.
    * @param numSamples The number of samples to write (length in time).
-   * @param dataInput A pointer to the data block.
+   * @param dataInput A vector of strings.
    * @param timestampsInput A pointer to the timestamps block.
    * @param controlInput A pointer to the control block data (optional)
    * @return The status of the write operation.
    */
   Status writeAnnotation(const SizeType& numSamples,
-                         const void* dataInput,
+                         const std::vector<std::string> dataInput,
                          const void* timestampsInput,
                          const void* controlInput = nullptr);
 
@@ -66,6 +66,6 @@ private:
   /**
    * @brief The number of samples already written per channel.
    */
-  SizeType m_samplesRecorded;
+  SizeType m_samplesRecorded = 0;
 };
 }  // namespace AQNWB::NWB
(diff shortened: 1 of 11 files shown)


src/nwb/misc/AnnotationSeries.hpp
```diff
@@ -0,0 +1,71 @@
...
+   * @return The status of the write operation.
+   */
+  Status writeAnnotation(const SizeType& numSamples,
+                         const std::vector<std::string> dataInput,
+                         const void* timestampsInput,
+                         const void* controlInput = nullptr);
+
+  DEFINE_FIELD(readData,
+               DatasetField,
+               std::any,
```
oruebel, line 61:
```suggestion
               std::string,
```

I believe this should be string instead of any, since the data for annotations are string data.

tests/testMisc.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::vector<std::string> expectedAnnotations = mockAnnotations;
+    expectedAnnotations.insert(expectedAnnotations.end(),
+                               mockAnnotations.begin(),
+                               mockAnnotations.end());
+    std::vector<std::string> dataOut(expectedAnnotations.size());
+
+    auto readAnnotationsData = io->readDataset(dataPath + "/data");
+    auto readAnnotationsDataTyped =
+        DataBlock<std::string>::fromGeneric(readAnnotationsData);
+    REQUIRE(readAnnotationsDataTyped.data == expectedAnnotations);
```
oruebel, line 60: Alternatively you could use the AnnotationSeries directly for read, e.g:

```
auto readDataWrapper = as->readData();
auto readAnnotationsDataTyped = readDataWrapper.values();
```

What you have here is fine. I just wanted to mention it, because I think using `as->readData` is more the typical workflow a user would take since it: a) uses the interface of the container class, rather than calling the io directly,  b) avoids having to specify the path., and c) when we have a data type specfied in the `DEFINE_FIELD` (e.g, here `std::string`), then we can even omit having to specify the data type as a template parameter, i.e., we can use `as->readData()` vs. `as->readData<std::string>()` (the latter is also fine, but not necessary). 

Similarly for `readTimestampsData` you could use `as->readTimestamps`.

stephprince: That makes sense, updated accordingly to match a more typical workflow.

stephprince:
@oruebel this is ready for you to re-review

oruebel (approved):

Issue #130, opened 2025-01-13
Read columns via `DynamicTable`


oruebel:
#85 adds core read capabilities. For `DynamicTable` specifically, it would be useful to have convenience methods to read `VectorData` columns that may also be specified dynamically and not be part of the schema. 

This is also related to #131 , which may partially address this issue or at least provide functionality for implementing a solution to this issue.

Currently, a user can get the path from `DynamicTable` and the `colNames` to construct the paths or use `findTypes` to search for columns in the `DynamicTable`. However, `DynamicTable` itself does not provide a way to create `ReadDataWrapper` objects for the `VectorData` columns it manages.

oruebel:
#142 adds the `RegisteredType::readField` method which supports reading of arbitrary subfields. To make this even more convenient we could add a `DEFINE_CHILD_FIELD` function to read fields that are itself instances of `RegisteredType` which would return the specific pointer type for the container (e.g., `VectorData`) instead of returning the generic `RegisteredType` pointer.

PR #142, opened 2025-01-16
Support reading of arbitrary fields 

### LINKED ISSUE #131: Support reading of custom fields via `Container`
#85 adds basic read capabilities. For `Container` it would be usefule to provide an access method to create `ReadDataWrapper` objects for attributes and datasets that are owned by the `Container` object but that to not have dedicated `DEFINE_FIELD` read method specified, e.g., to read extra user-defined fields or fields in the schema that may not yet be covered by AqNWB. Currently a user can read those manually by defining explicit paths and creating `ReadDataWrapper` objects manually or by reading directly via the io backend.

### LINKED ISSUE #133: Enhance direct unit tests for basic and private read functions
This is a follow-up to #85

- [X] Add unit test for ``src/Utils.cpp ,  mergePaths()`` function (See #137 )
- [X] Add unit tests for new private ``HDF5IO`` internal functions. These may not require separate dedicated tested, but may be covered via the tests of the public  core interfaces functions. However, we should check that they are indeed covered. **NOTE:** I checked with Cline and added another test in #139
   - [X] ``HDF5IO:: readDataHelper`` (note, there are two variants of this function)
   - [X] ``HDF5IO::readStringDataHelper`` (note there are two variants of this function)
   - [X] ``HDF5IO::getAttribute``
- [X] Add unit test for ``BaseIO::findTypes`` (possibly as part of the HDF5IO tests) (See #139)
- [x] Check unit test coverage for ``RegisteredType``. This should be largely covered by the other tests, but we should check. (See #139)
- [x] Check unit test coverage for ``ReadIO``. This should be largely covered by the other tests, but we should check.

oruebel:
This PR is to prepare for #130 

*  Changed the `getGroupObjects` method to `getStorageObjects` to allow for more general lookup of groups, datasets, and attributes. The current method is limited to look-up of groups within a group. The updated `getStorageObjects` now supports: 1) lookup starting from datasets and groups and 2) retrieval of groups, datasets, and attributes, 3) allows the user to specify what types of storage objects to retrieve. 
* Fixed an issue in the `findTypes` function, which previously only looked at `Groups` so typed objects that were datasets (e.g., `VectorData`) would have been ignored. 
* Added unit tests for `findTypes` and missing tests for `RegisteredType` and `HDF5IO`
* Added new `RegisteredType::readField` function for reading arbitrary dataset and attribute fields
* Added second variant of  the `RegisteredType::readField` method to allow reading of arbitrary fields that are itself instances of `RegisteredType`, such as, reading `VectorData` from a `DynamicTable` or reading `ElectricalSeries` from an `NWBFile`

Fix #133 
Fix #131

### DIFF
--- a/tests/testHDF5IO.cpp
+++ b/tests/testHDF5IO.cpp
@@ -128,21 +128,21 @@ TEST_CASE("createGroup", "[hdf5io]")
   }
 }
 
-TEST_CASE("getGroupObjects", "[hdf5io]")
+TEST_CASE("getStorageObjects", "[hdf5io]")
 {
   // create and open file
-  std::string filename = getTestFilePath("test_getGroupObjects.h5");
-  IO::HDF5::HDF5IO hdf5io(filename);
+  std::string filename = getTestFilePath("test_getStorageObjects.h5");
+  AQNWB::IO::HDF5::HDF5IO hdf5io(filename);
   hdf5io.open();
 
   SECTION("empty group")
   {
     hdf5io.createGroup("/emptyGroup");
-    auto groupContent = hdf5io.getGroupObjects("/emptyGroup");
+    auto groupContent = hdf5io.getStorageObjects("/emptyGroup");
     REQUIRE(groupContent.size() == 0);
   }
 
-  SECTION("group with datasets and subgroups")
+  SECTION("group with datasets, subgroups, and attributes")
   {
     hdf5io.createGroup("/data");
     hdf5io.createGroup("/data/subgroup1");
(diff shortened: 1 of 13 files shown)


oruebel:
@stephprince sorry in case this caused confusion, but I needed the changes from #139 here for testing, and since #139 was not merged yet I combined it with this PR so I could run the tests. This PR is now ready for review.

tests/testHDF5IO.cpp
```diff
@@ -128,21 +128,21 @@ TEST_CASE("createGroup", "[hdf5io]")
   }
 }
 
-TEST_CASE("getGroupObjects", "[hdf5io]")
+TEST_CASE("getStorageObjects", "[hdf5io]")
```
stephprince, line 131: Two other test cases that I don't think are covered are providing an invalid path or a path to an attribute - I think those would return empty groupContent but you could add the tests if you think it's needed

stephprince (dismissed):
Looks good to me! Added some minor suggestions.

tests/testHDF5IO.cpp
oruebel: Done. I added the additional test case in #[19fecb4](https://github.com/NeurodataWithoutBorders/aqnwb/pull/142/commits/19fecb4225c58b70e8448124878a537fd8b72d92)

oruebel:
@stephprince can you please re-approve. I committed your suggestions and added the test case you had suggested in [19fecb4](https://github.com/NeurodataWithoutBorders/aqnwb/pull/142/commits/19fecb4225c58b70e8448124878a537fd8b72d92)

stephprince (approved):

PR #143, opened 2025-01-19
Support reading of arbitrary RegisteredTypes 

### LINKED ISSUE #130: Read columns via `DynamicTable`

oruebel:
Fix #130 

Support reading typed columns of a DynamicTable and as a result support more broadly reading of RegisteredType objects via predefined functions via the new `DEFINE_REGISTERED_FIELD` macro, much like the existing `DEFINE_FIELD` macro.


oruebel:
@stephprince I made this a separate PR from #142, because I: 1) didn't want to make that PR more complicated, 2) allow #142 to be merged independently, and 3) felt that the changes in this PR are a bit more fundamental and may require a closer look and maybe some discussion.

oruebel:
@stephprince I took another look at it. The overall approach for reading types is working and is consistent with how we define the read of arrays/attributes. From a user perspective, I think having `Data` and `VectorData` with the template parameter should be fine. By default, `VectorData` has default value of `VectorData<std::any>` so the user can still use `VectorData` as before and delay specifying the data type until later when actually reading values. Also, the default behavior on read is still `VectorData<std::any>` because in the NWB file only specify `neurodata_type=VectorData`. However, for columns where we have aknown data type, e.g. `location`, setting the template parameter `VectorData<std::string>` in the `DEFINE_REGISTERED_FIELD` macro allows us to provide read methods that already specify the data type (similar to how we do this for reading arrays). The advantage is that the user then doesn't need to set the data type by hand. 

Long story short, after taking another look, I think we can merge this PR as is. If we want to try a different approach for `Data` and `VectorData` to try and remove the need for a template parameter, then I think it would be Ok to move that discussion to a separate issue/PR.

stephprince (approved):
Looks good to me!

PR #144, opened 2025-01-19
Allow setting of the session start time and timestamp reference time

### LINKED ISSUE #8: Add option for acquisition system to supply session start and reference times
`session_start_time` and `timestamps_reference_time` are set in `NWBFile.cpp` to the current time. 

```c++
  io->createStringDataSet("/session_start_time", time);
  io->createStringDataSet("/timestamps_reference_time", time);
```

We may want to add the option for the acquisition system to supply those times as parameters.

oruebel:
Fix #8 

This PR is not critical, just taking care of a minor issue that was assigned to me

### DIFF
--- a/src/nwb/NWBFile.cpp
+++ b/src/nwb/NWBFile.cpp
@@ -49,15 +49,6 @@ NWBFile::~NWBFile() {}
 
 Status NWBFile::initialize(const std::string& identifierText,
                            const std::string& description,
-                           const std::string& dataCollection)
-{
-  std::string time = getCurrentTime();
-  return this->initialize(
-      identifierText, description, dataCollection, time, time);
-}
-
-Status NWBFile::initialize(const std::string& identifierText,
-                           const std::string& description,
                            const std::string& dataCollection,
                            const std::string& sessionStartTime,
                            const std::string& timestampsReferenceTime)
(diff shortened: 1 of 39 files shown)


src/nwb/NWBFile.cpp
```diff
@@ -50,14 +50,41 @@ NWBFile::~NWBFile() {}
...
+  std::string time = getCurrentTime();
+  return this->initialize(
+      identifierText, description, dataCollection, time, time);
+}
+
+Status NWBFile::initialize(const std::string& identifierText,
+                           const std::string& description,
+                           const std::string& dataCollection,
+                           const std::string& sessionStartTime,
+                           const std::string& timestampReferenceTime)
```
stephprince, line 63:
```suggestion
                           const std::string& timestampsReferenceTime)
```

Very minor and I don't think we need to change necessarily, but it's defined as `timestamps_reference_time` with an s in the file if we want to be consistent with naming

src/nwb/NWBFile.hpp
```diff
@@ -67,11 +67,31 @@ class NWBFile : public Container
...
+   * @param sessionStartTime ISO formatted time string with the session start
+   * time
+   * @param timestampReferenceTime ISO formatted time string with the timestamp
+   * reference time
+   */
+  Status initialize(const std::string& identifierText,
+                    const std::string& description,
+                    const std::string& dataCollection,
+                    const std::string& sessionStartTime,
+                    const std::string& timestampReferenceTime);
```
stephprince, line 93: I am wondering in this case if it would be clearer to provide default arguments for `sessionStartTime` and `timestampReferenceTime` instead of overloading the initialize function since they are just additional arguments and not different types for the same input? 

We could then add a comment in the documentation about the default option being the current time if none is provided.

stephprince (commented):
Looks good to me! Added a couple of comments.

src/nwb/NWBFile.hpp
oruebel: What should the default argument be? We can't use the current time directly as a default since that is not known at compile time. I guess we could use empty string as the default and say in the docs, if it is empty then we'll use the current time.

src/nwb/NWBFile.cpp
oruebel: Done

src/nwb/NWBFile.hpp
oruebel: Ok, I updated the code to use empty string as default and remve the overloading of the method as you suggested.

oruebel:
@stephprince I implemented the code changes you had suggested.

stephprince (approved):

PR #148, opened 2025-01-24
Enhance unit test coverage

oruebel:
Most of this PR was done with Cline. Add separate unit tests for select classes to enhance overall coverage. Add unit tests for:

- [X] `HDF5Recording`
- [X] `Channel`

### DIFF
--- a/tests/testChannel.cpp
+++ b/tests/testChannel.cpp
@@ -1,0 +1,189 @@
+#include <catch2/catch_approx.hpp>
+
+#include "Channel.hpp"
+#include "testUtils.hpp"
+
+TEST_CASE("Test Channel Construction and Basic Operations", "[channel]")
+{
+  SECTION("Default constructor parameters")
+  {
+    Channel ch("test_channel", "test_group", 1, 2, 3);
+    REQUIRE(ch.getName() == "test_channel");
+    REQUIRE(ch.getGroupName() == "test_group");
+    REQUIRE(ch.getGroupIndex() == 1);
+    REQUIRE(ch.getLocalIndex() == 2);
+    REQUIRE(ch.getGlobalIndex() == 3);
+    REQUIRE(ch.getConversion()
+            == Catch::Approx(0.05e-6f).epsilon(
+                0.001));  // bitVolts/conversion = 0.05/1e6
+    REQUIRE(ch.getSamplingRate() == Catch::Approx(30000.f).epsilon(0.001));
+    REQUIRE(ch.getBitVolts() == Catch::Approx(0.05f).epsilon(0.001));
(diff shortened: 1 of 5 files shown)


oruebel:
@stephprince this is ready for review

tests/testChannel.cpp
```diff
@@ -0,0 +1,189 @@
...
+    REQUIRE(ch.getLocalIndex() == 2);
+    REQUIRE(ch.getGlobalIndex() == 3);
+    REQUIRE(ch.getConversion() == Catch::Approx(0.1f / 2e6f).epsilon(0.001));
+    REQUIRE(ch.getSamplingRate() == Catch::Approx(44100.f).epsilon(0.001));
+    REQUIRE(ch.getBitVolts() == Catch::Approx(0.1f).epsilon(0.001));
+    const auto& actualPos = ch.getPosition();
+    REQUIRE(actualPos[0] == Catch::Approx(pos[0]).epsilon(0.001));
+    REQUIRE(actualPos[1] == Catch::Approx(pos[1]).epsilon(0.001));
+    REQUIRE(actualPos[2] == Catch::Approx(pos[2]).epsilon(0.001));
+    REQUIRE(ch.getComments() == "test comment");
```
stephprince, line 55: I think we can use the `Catch::Matchers::RangeEquals` with an approximate comparator to simplify our comparison of vectors/arrays of floats throughout these tests. 

```suggestion
    auto approxComparator = [](float a, float b) { 
        return a == Catch::Approx(b).epsilon(0.001);   # define once or in testUtils
    };
    REQUIRE_THAT(actualPos, Catch::Matchers::RangeEquals(pos, approxComparator));
```

tests/testChannel.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  SECTION("Set name")
+  {
+    ch.setName("new_channel");
+    REQUIRE(ch.getName() == "new_channel");
+  }
+}
+
+TEST_CASE("Test Channel Copy and Move Operations", "[channel]")
+{
```
stephprince, line 87: Is testing copy/move operations something we want to add for all classes that have these types of constructors? That seems like it might be overkill, but maybe there is a reason it is particularly important for the Channel class? If we were going to test on a subset of classes it might be particularly relevant to the Container/Data classes since those are moved during the recording workflow.

oruebel: > That seems like it might be overkill,

I don't think it is strictly necessary. I mainly kept them because it helps with coverage but if you prefer to remove them to avoid bloat in the tests then that's fine with me.

stephprince (approved):
Looks good to me - thanks!

PR #149, opened 2025-01-31
Update handling of colnames

### LINKED ISSUE #107: Adding columns to DynamicTable results in invalid colNames
Currently, ``DynamicTable`` created the ``colNames`` attribute in its ``initalize`` method here:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.cpp#L26

Adding columns via ``DynamicTable.addColumn``, however, does not update the ``colNames``. 

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/hdmf/table/DynamicTable.cpp#L29-L47

As such, we must set ``colNames`` before calling ``DynamicTable`` as otherwise the ``colNames`` will not include any added columns, and is hence invalid.

``ElectrodeTable`` works around this issue, by: 1) fixing``colNames`` on construction, and 2) adding a ``finalize`` method in which all columns are added at once, rather than allowing columns to be added individually via ``addColumn``

https://github.com/NeurodataWithoutBorders/aqnwb/blob/5c37c400e84952f4e560b82223a0d777aa70a234/src/nwb/file/ElectrodeTable.cpp#L51-L65

However, if a user would call ``addColumn`` on ``ElectrodeTable` to add other columns it would result in a bad `colNames`. 

## Possible solutions

* **1)** Add a ``DynamicTable.finalize`` method and set ``colNames`` at the end rather than in ``initialize`` so that we can update ``colNames`` when adding new columns
   * The disadvantage of this is that the ``DynamicTable`` is not valid until ``finalize`` is called. This is not too bad for metadata tables that are being created and finalized at the beginning, however, this is problematic for tables that we may want to acquire data into, e.g,. ``TimeIntervals`` or ``PlaneSegmentation``. 
* **2)** Update ``colNames`` dynamically, i.e., overwrite the attribute each time a new column is being added
   * The disadvantage of this is that we potentially update the attribute repeatedly but provides maximum flexibility  
**3)** Add a ``DynanmicTable.finalizeColumns`` method to finish defining all columns to only allow adding new rows but not new columns. 
   * This seems like a possible compromise and would be consistent with the behavior of `SWMR` mode in the `HDFIO` backend. I.e., we would call this function before enabling `SWMR` mode. We could potentially generalize this, and have a `Container.finalizeFields` function so that we could all `finalizeFields` for all `Containers` in the `NWBFile`

We could ``finalizeColumns``

### LINKED ISSUE #129: Move `colNames` from `DynamicTable` constructor to `initialize`
To ensure `DynamicTable` can be created consistent with `RegisteredClass` interface, the `colNames` field should be removed from the constructor and instead be moved to the `initialize` function. See also related discussions in #85 here  https://github.com/NeurodataWithoutBorders/aqnwb/pull/85/

This is also in part related to #107, which discusses handling of `colNames` in a more dynamic settings where columns are being added incrementally.

### LINKED ISSUE #150: Update
Writing an attribute with an array of strings currently writes as fixed-length strings:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/00a529cee1e074510c9cf3724ea419bff26a834a/src/io/hdf5/HDF5IO.cpp#L684-L697

It would be nice to write string data as variable length strings instead.

oruebel:
This PR updates the handling of the `DynamicTable.colNames` to: 1) avoid the need for specifying `colNames` at the beginning and 2) to allow us to dynamically add columns to a `DynamicTable` (even one that was read from a file)

# Related issues
Fix #107
Fix #150  
Fix #129 

## Fixed writing of variable-length string attributes and `VectorData` columns:
- [X] Fixed writing of Attributes that are arrays of variable-length 
    - [X] Write string array attributes as variable length string arrays instead of fixed length string arrays
    - [X] Updated `BaseIO.createAttribute` for string arrays to explicitly allow us to force overwrite of existing attributes and updated `HDF5IO.createAttribute` accordingly. HDF5 does not support expanding attributes in the same way it does for datasets so we need to delete the existing attribute and overwrite it with a new one with the same name. 
    - [X] Fixed writing of scalar string attributes to be written as actual scalar strings. Previously  `HDF5IO.createAttribute` for scalar strings transformed the scalar string to an array of strings and wrote it using the array method. This PR implements the function to write scalar strings. 
- [X] Fixed writing of variable-length strings columns via `DynamicTable.add_column` to avoid unnecessary iteration and use `writeDataBlock` directly
- [X] Fixed data types for string columns in `ElectrodeTable` to use variable length strings instead of fixed length strings

## Improved handling of `Status`
- [X] Return the `Status` for `DynamicTable` functions that perform write operations (e.g,. `initialize`, `finalize`, `add_column` etc.) as well as relevant other classes, e.g., `Data`, `VectorData`, `Container` ...
- [X]  Defined ``&&`` and ``||`` operator for the ``Status`` enum to simplify combining multiple `Status` checks 

## Updated handling of `colNames` to correctly handle dynamic adding of columns
- [X] Delayed writing of `colNames` to the end by adding `DynamicTable.finalize` akin to the existing `ElectrodeTable.finalize`. This allows for columns to be added to the `DynamicTable`
    - [X] Updated `DynamicTable.finalize` to overwrite column names to correctly support appending of columns to existing tables
- [X] Updated `DynamicTable` constructor to read the existing column names if they exist from the file so that we can append new columns to existing tables

## Updated unit tests 
- [X] Updated unit tests for `HDF5IO.createAttribute`
- [X] Added new unit tests for `DynamicTable` 
- [X] Enhanced `ElectrodeTable` unit tests to appropriately check for the status of write operations and fix existing failing tests due to missing `Device` and `ElectrodeGroup`


oruebel:

src/io/hdf5/HDF5IO.cpp
```diff
@@ -683,17 +683,82 @@ Status HDF5IO::createAttribute(const std::string& data,
 
 Status HDF5IO::createAttribute(const std::vector<std::string>& data,
                                const std::string& path,
-                               const std::string& name)
+                               const std::string& name,
+                               const bool overwrite)
```
stephprince, line 687: Looks good, I am just wondering if there was a specific reason to add the overwrite functionality on this function instead of the version of the `createAttribute` function defined below:

```c++
Status HDF5IO::createAttribute(const std::vector<const char*>& data,
                               const std::string& path,
                               const std::string& name,
                               const SizeType& maxSize)
```
Since that version is used during both single string and string array attribute creation and performs similar checks for the attribute already existing?

src/io/hdf5/HDF5IO.hpp
```diff
@@ -152,18 +152,22 @@ class HDF5IO : public BaseIO
...
    */
   Status createAttribute(const std::vector<std::string>& data,
                          const std::string& path,
-                         const std::string& name) override;
+                         const std::string& name,
+                         const bool overwrite = false) override;
 
   /**
-   * @brief Creates a string array attribute at a given location in the file.
+   * @brief Creates a fixed-length string array attribute at a given location in
```
stephprince, line 169: From my understanding this function writes variable length string arrays:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/d69d55c8f399f22214aa279838a8a4daa3e87b91/src/io/hdf5/HDF5IO.cpp#L780

However, I think it is maybe unclear since a maximum size is accepted, but this parameter is then overwritten. I'm not sure why that is the case....

src/nwb/file/ElectrodeTable.hpp
```diff
@@ -40,41 +40,27 @@ class ElectrodeTable : public DynamicTable
    * Initializes the ElectrodeTable by creating NWB related attributes and
    * adding required columns.
    *
-   *  @param description The description of the table (default: "metadata about
+   * @param description The description of the table (default: "metadata about
```
stephprince, line 43:
```suggestion
   * @param description The description of the table (default: "metadata about extracellular electrodes")
```

stephprince (commented):
Looks great! Thanks. I just had a question about the overwriting of variable length strings but all of the electrodes table and column changes look good to me.

src/io/hdf5/HDF5IO.hpp
oruebel: I was digging around a bit, and I believe the call to `H5type.setSize(H5T_VARIABLE);` is actually a but. The type is being created as a fixed type here ``StrType H5type(PredType::C_S1, maxSize);`` and changing the size after the fact to variable may not actually work. If we want variable length strings here then we should use "StrType H5type(PredType::C_S1, H5T_VARIABLE);"  directly. We decided to remove the fixed-length string variant in liu of the variable length string function.

src/io/hdf5/HDF5IO.cpp
oruebel: We decided to remove the fixed length string variant and use the variable length string variant instead. I'll submit and update for this shortly.

oruebel: I've removed the `creataAttribute` variant for fixed length strings https://github.com/NeurodataWithoutBorders/aqnwb/pull/149/commits/9217da351d09e1582b1f660b09c0461684436597

I've also updated `HDF5IO.createAttribute` for scalar strings to write the data as scalar strings instead of as an array of length 1.

oruebel:
@stephprince please take another look. I believe I've addressed your suggestion.

src/io/hdf5/HDF5IO.cpp
```diff
@@ -731,28 +774,48 @@ Status HDF5IO::createAttribute(const std::vector<const char*>& data,
...
     } else {
-      DataSpace attr_dataspace;
-      SizeType nStrings = data.size();
-      if (nStrings > 1) {
-        dims[0] = nStrings;
-        attr_dataspace = DataSpace(1, dims);
-      } else
-        attr_dataspace = DataSpace(H5S_SCALAR);
-      attr = loc->createAttribute(name, H5type, attr_dataspace);
+      attr_dataspace = DataSpace(H5S_SCALAR);
```
stephprince, line 791: If I understand correctly, since we now have a separate function that writes a single string, we would not want this else statement since the user may want to write an array of length 1 instead of a scalar?

oruebel: Good catch. Yes, that should be an array of size 1 instead.

oruebel: I fixed this in https://github.com/NeurodataWithoutBorders/aqnwb/pull/149/commits/044e2d3431c038cad63d5dbf23b43949e206063c and also added a corresponding unit test to make sure we can write both attributes that are scalar string and attributes that are an array of strings with just a single value.

oruebel:
@stephprince I resolved all merge conflicts with main and fixed the issue with `createAttribute` in https://github.com/NeurodataWithoutBorders/aqnwb/commit/044e2d3431c038cad63d5dbf23b43949e206063c

stephprince (approved):

PR #156, opened 2025-02-15
Fix ElectricalSeries->electrodes dataset write and add tests

stephprince:
The `ElectricalSeries->electrodes` dataset was being written incorrectly due to an incorrect data type. I updated the data type and added tests for write/read of this `ElectricalSeries->electrodes` dataset to avoid this issue in the future.

### DIFF
--- a/src/nwb/ecephys/ElectricalSeries.hpp
+++ b/src/nwb/ecephys/ElectricalSeries.hpp
@@ -109,6 +109,17 @@ public:
                Base unit of measurement for working with the data. 
                This value is fixed to volts)
 
+  DEFINE_FIELD(readElectrodes,
+               DatasetField,
+               int,
+               "electrodes",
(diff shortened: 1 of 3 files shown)


oruebel (approved):

src/nwb/ecephys/ElectricalSeries.hpp
```diff
@@ -109,6 +109,17 @@ class ElectricalSeries : public TimeSeries
                Base unit of measurement for working with the data. 
                This value is fixed to volts)
 
+  DEFINE_FIELD(readElectrodes,
+               DatasetField,
+               int,
+               "electrodes",
+               The electrodes that generated this electrical series.)
```
oruebel, line 116:
```suggestion
               The indices of the electrodes that generated this electrical series.)
```

PR #158, opened 2025-02-16
Add support for reading reference attributes

### LINKED ISSUE #157: Implement read for references stored as attributes

oruebel:
- [X] Added `readReferenceAttribute` to `BaseIO` and `HDF5IO` to implement reading of attribute references and added unit tests
- [X] Added `DEFINE_REFERENCED_REGISTERED_FIELD` macro to `RegisteredTypes.hpp` to allow direct read of registered objects that are referenced by an attributed, e.g., to read the `ElectrodesTable` from the `ElectricalSeries` or to -read `VectorData` from a `VectorIndex`
- [X] Updated `ElectricalSeries` to add definition of `readElectrodesTable` function for reading the `electrodes/table` attribute and return the `ElectrodeTable` object and added a unit test

Fix #157


oruebel:

docs/pages/devdocs/registered_types.dox
```diff
@@ -230,6 +230,16 @@
  * DEFINE_REGISTERED_FIELD(getData, DynamicTable, "my_table", My data table)
  * @endcode
  *
+ * @section use_the_define_referenced_registered_field_macro How to use the DEFINE_REFERENCED_REGISTERED_FIELD macro
+ *
+ * The \ref DEFINE_REFERENCED_REGISTERED_FIELD macro works exactly like the 
+ * \ref DEFINE_REGISTERED_FIELD macro, but the underlying data is an attribute that 
+ * stores a reference to an instances of a specific subtype of \ref AQNWB::NWB::RegisteredType "RegisteredType"
+ * rather than the instance of the object directly. I.e., ``fieldPath`` here is the
+ * relative path to attribute that stores the reference, rather than the relative path
```
stephprince, line 239:
```suggestion
 * relative path to the attribute that stores the reference, rather than the relative path
```

src/nwb/RegisteredType.hpp
```diff
@@ -427,5 +427,55 @@ class RegisteredType
     return nullptr; \
   }
 
+/**
+ * @brief Defines a lazy-loaded accessor function for reading fields that are
+ * RegisteredTypes that are linked to by a reference attribute
+ *
+ * This macro generates a function that returns the approbriate subtype of
+ * RegisteredType, e.g., to read VectorData from a DynamicTable or a
+ * TimeSeries from and NWBFile.
```
stephprince, line 436:
```suggestion
 * This macro generates a function that returns the appropriate subtype of
 * RegisteredType, e.g., to read VectorData from a DynamicTable or a
 * TimeSeries from an NWBFile.
```

src/nwb/ecephys/ElectricalSeries.hpp
```diff
@@ -109,17 +110,27 @@ class ElectricalSeries : public TimeSeries
...
                "electrodes/description",
                The electrodes that generated this electrical series.)
+
+  DEFINE_REFERENCED_REGISTERED_FIELD(
+      readElectrodesTable,
+      ElectrodeTable,
+      "electrodes/table",
+      Table with all extracellular electrodes retrieved from the reference to
+              the electrode table stored in the `electrodes
+          / table` attribute)
```
stephprince, line 132:
```suggestion
     The electrodes table retrieved from the object referenced in the 
      `electrodes/table` attribute.)
```

I found the description a little hard to understand, tried to clarify

src/io/hdf5/HDF5IO.cpp
```diff
@@ -301,6 +301,59 @@ std::vector<std::string> HDF5IO::readStringDataHelper(
...
+        "HDF5IO::readReferenceAttribute, attribute is not a reference.");
+  }
+
+  // Read the reference
+  hobj_ref_t ref;
+  attribute.read(dataType, &ref);
+
+  // Dereference the reference to get the HDF5 object ID
+  hid_t obj_id =
+      H5Rdereference2(attribute.getId(), H5P_DEFAULT, H5R_OBJECT, &ref);
```
stephprince, line 328: Note that this function is [deprecated as of HDF5 1.12](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_r.html#ga9b09586f7b6ec708434dd8f95f58a9b7), however we currently support HDF5 > 1.10 so I do not think we can use the replacement functions yet.

src/io/hdf5/HDF5IO.cpp
```diff
@@ -301,6 +301,59 @@ std::vector<std::string> HDF5IO::readStringDataHelper(
...
+    throw std::runtime_error(
+        "HDF5IO::readReferenceAttribute, failed to get object name size.");
+  }
+
+  std::vector<char> obj_name(static_cast<size_t>(buf_size));
+  if (H5Iget_name(obj_id, obj_name.data(), static_cast<size_t>(buf_size)) < 0) {
+    H5Oclose(obj_id);
+    throw std::runtime_error(
+        "HDF5IO::readReferenceAttribute, failed to get object name.");
+  }
```
stephprince, line 347: Under what circumstances would these errors occur that it fails to get the name of the dereferenced object? Should we add any additional tests for those

tests/testHDF5IO.cpp
```diff
@@ -636,10 +636,47 @@ TEST_CASE("HDF5IO; create attributes", "[hdf5io]")
...
+      hdf5io.createReferenceAttribute(
+          "/referenceTargetDataset", "/data", "datasetRefAttr");
+      REQUIRE(hdf5io.attributeExists("/data/datasetRefAttr"));
+
+      std::string resolvedPath =
+          hdf5io.readReferenceAttribute("/data/datasetRefAttr");
+      REQUIRE(resolvedPath == "/referenceTargetDataset");
+    }
+
+    // Test reading non-existent reference attribute
```
stephprince, line 673: Can you add an additional test for using `readReferenceAttribute` on an attribute that is not a reference?

stephprince (commented):
Added a few comments but otherwise looks good to me!

src/io/hdf5/HDF5IO.cpp
oruebel: Interesting. What approach are you suggesting we use to deal with this, we could: 

1. just add a TODO comment to update when upgrading to HDF5 >1.12, or
2. have and ``#ifdef`` logic that depends on the HDF5 version to use the new variant for newer version and the old way for older versions, or
3. Update to HDF5 >1.12 and update the code to use the new logic

src/io/hdf5/HDF5IO.cpp
oruebel: I'm honestly not sure what might trigger those errors. I think those checks are mostly useful as extra safeguards, but I don't think they are strictly necessary.

tests/testHDF5IO.cpp
oruebel: Sure, will do

src/io/hdf5/HDF5IO.cpp
oruebel: I.e., I'm not sure how to really test theses. This sort of thing I think would mainly happen if there was some outside intervention, e.g., deleting the file in the middle of running the code. What is your preference: a) remove the checks, or b) leave the checks as safeguards but without tests (and add a comment)

stephprince: Leaving is fine to me! I was more unsure when exactly they would be triggered.

oruebel: Ok. I added a comment in the code and excluded those lines of lcov code coverage since we can't easily test this.

src/io/hdf5/HDF5IO.cpp
oruebel: I added a TODO note so that we can update when we move to a newer version of HDF5

oruebel:
@stephprince thanks for the comments. I believe I have now addressed all your suggestions.

stephprince (approved):
Looks good to me - thanks!

PR #161, opened 2025-02-20
Move ElectrodesTable creation

### LINKED ISSUE #155: Electrodes indices can be out of range
`ElectricalSeries.initialize` currently uses the global electrode indices when creating the electrodes dataset:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/d234fa004a82ce0af5009ef18428cd5089aa7f71/src/nwb/ecephys/ElectricalSeries.cpp#L45-L52

If all channels detected by the acquisition system are being actively recorded from, this works fine. However if a subset of all available electrode channels are actively recording data, then our convenience methods `nwbfile->createElectricalSeries` and `nwbfile->createSpikeEventSeries` can cause potential issues, because they use the same `channelVector` to create the electrodes table and create any datasets being recorded to. As a result, the global index can be larger than the actual length of the electrodes table. This came up when I was testing the latest AqNWB with the OpenEphys plugin neuropixel simulator, where a subset of channels can be recorded from at anytime.

Should the electrodes table always contain all channels detected by the acquisition system (i.e. the full range of the globalIndex)? I think this is the safest route and is what OpenEphys currently does. If so, then should the electrodes table creation be separated out from the at `createElectricalSeries` or `createSpikeEventSeries` convenience methods? These methods currently detect whether the table exists and create it if needed.

stephprince:
Fix #155 by moving the creation of ElectrodesTable to a separate convenience function.

As part of these changes, I added some additional tests, a helper function for getting the size of a dataset, and a check for creating ElectricalSeries with electrodes indices outside of the possible range of the ElectrodesTable.


docs/pages/userdocs/workflow.dox
```diff
@@ -14,12 +14,14 @@
  *    used for managing \ref AQNWB::NWB::Container "Container" objects for storing recordings.
  * 3. Create the \ref AQNWB::NWB::NWBFile "NWBFile" object used for managing and creating NWB
  *    file contents.
- * 4. Create the \ref AQNWB::NWB::Container "Container" objects (e.g.,
+ * 4. Create the \ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" object used for saving 
+ *    electrodes information.
```
oruebel, line 18:
```suggestion
 * 4. Create the recording metadata stored in ``\general`` in NWB, e.g., the  
 *    \ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" object used for 
 *    saving electrodes information. 
```

docs/pages/userdocs/workflow.dox
```diff
@@ -52,7 +54,22 @@
...
+ *
+ * Next, constructs the \ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" object, 
+ * using the `recordingArrays` as an input. Note that in many cases, you will be recording from 
+ * all electrodes detected by the device and this `recordingArrays` structure will be the same as
+ * is used to create the \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries". However in some cases
+ * (e.g. Neuropixels), you may be recording from a subset of all available electrodes. In this case,
+ * you should create the ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" with all electrodes
+ * detected by the acquisition system, not just those being actively recorded from. This approach is
+ * critical for mapping information about which \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries"
+ * were recorded from which electrodes.
```
oruebel, line 67:
```suggestion
 * \subsection create_recmeta 4. Create the recording metadata 
 * \subsubsection create_recmeta_ecephys Create the extracellular recording metadata
 * Next, constructs the \ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" object, 
 * using the `recordingArrays` as an input. Note that in many cases, you will be recording from 
 * all electrodes detected by the device and this `recordingArrays` structure will be the same as
 * is used to create the \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries". However in some cases
 * (e.g. Neuropixels or when using multiple probees), you may be recording from a subset of 
 * all available electrodes as part of a single \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries". In this case,
 * you should create the ref AQNWB::NWB::ElectrodesTable "ElectrodesTable" with all electrodes
 * detected by the acquisition system, not just those being actively recorded from. This approach is
 * critical for mapping information about which \ref AQNWB::NWB::ElectricalSeries "ElectricalSeries"
 * were recorded from which electrodes.
```

src/io/BaseIO.hpp
```diff
@@ -441,6 +441,13 @@ class BaseIO
   virtual std::unique_ptr<BaseRecordingData> getDataSet(
       const std::string& path) = 0;
 
+  /**
+   * @brief Returns the size of the dataset for each dimension.
+   * @param path The location in the file of the dataset.
+   * @return The dataset size.
+   */
+  virtual std::vector<SizeType> getDatasetSize(const std::string path) = 0;
```
oruebel, line 449: - It would be nice to have this function support both datasets and attributes. In that case, I would rename it to `getStorageObjectShape`.  
- It would be nice to also add the following function to `ReadDataWrapper` to make this new function convenient to use on read

```  
inline std::vector<SizeType> getShape() const
{
   return m_io->getShape(m_path);
}

- We should be consistent in how we use `shape` and `size`. I think we should use `shape` to refer to the sizes along all dimensions and `size` to refer to the total number of elements.

File: src/nwb/NWBFile.cpp [hunk 4]
```diff
@@ -28,7 +28,7 @@ constexpr SizeType CHUNK_XSIZE =
 constexpr SizeType SPIKE_CHUNK_XSIZE =
     8;  // TODO - replace with io settings input
 
-std::vector<SizeType> NWBFile::emptyContainerIndexes = {};
+std::vector<SizeType> NWBFile::m_emptyContainerIndexes = {};
```
oruebel, line 31: Maybe `Types.h` is a better place for this or alternatively, I think we can also just explicitly use "std::vector<SizeType>()" as default argument (at least it seems that `m_emptyContainerIndexes` is only used to set default arguments of functions, i.e., we could just write:

```
Status createAnnotationSeries(
      std::vector<std::string> recordingNames,
      RecordingContainers* recordingContainers = nullptr,
      std::vector<SizeType>& containerIndexes = std::vector<SizeType>());
```

stephprince: If I remember correctly, I think I set this up this way because you cannot pass a temporary empty initializer list as a default argument to a non constant reference parameter. Creating the empty vector in advance was a solution for that. Another potential option is to change the parameter to an rvalue reference, but then I think there might be additional complications in terms of transferring ownership.

src/io/BaseIO.hpp
stephprince: > It would be nice to have this function support both datasets and attributes. In that case, I would rename it to getStorageObjectShape.
It would be nice to also add the following function to ReadDataWrapper to make this new function convenient to use on read

That makes sense, I will add functionality for attributes and use `shape` to refer to sizes along all dimensions

stephprince:
Based on our discussion, I made `containerIndexes` a required argument and removed the static `m_emptyContainerIndexes` variable. I think `recordingContainers` should likely also be required as an input for the user so I removed the default value for that parameter as well.

These changes will now also fix #162, if we want to take a different approach or discuss that particular issue more I can move the changes to `getStorageObjectsShape` to a separate PR.

stephprince:
@oruebel this is ready for you to re-review

docs/pages/userdocs/workflow.dox
```diff
@@ -14,12 +14,15 @@
  *    used for managing \ref AQNWB::NWB::Container "Container" objects for storing recordings.
  * 3. Create the \ref AQNWB::NWB::NWBFile "NWBFile" object used for managing and creating NWB
  *    file contents.
- * 4. Create the \ref AQNWB::NWB::Container "Container" objects (e.g.,
+ * 4. Create the recording metadata stored in ``\general`` in NWB, e.g., the
```
oruebel, line 17:
```suggestion
 * 4. Create the recording metadata stored in ``/general`` in NWB, e.g., the  
```

src/io/BaseIO.hpp
```diff
@@ -441,6 +441,14 @@ class BaseIO
   virtual std::unique_ptr<BaseRecordingData> getDataSet(
       const std::string& path) = 0;
 
+  /**
+   * @brief Returns the size of the dataset for each dimension.
+   * @param path The location in the file of the dataset.
+   * @return The dataset shape.
```
oruebel, line 447:
```suggestion
   * @brief Returns the size of the dataset or attribute for each dimension.
   * @param path The location of the dataset or attribute in the file
   * @return The shape of the dataset or attribute.
```

src/io/hdf5/HDF5IO.hpp
```diff
@@ -268,6 +268,13 @@ class HDF5IO : public BaseIO
   std::unique_ptr<IO::BaseRecordingData> getDataSet(
       const std::string& path) override;
 
+  /**
+   * @brief Returns the size of the dataset for each dimension.
+   * @param path The location in the file of the dataset.
+   * @return The dataset shape.
+   */
```
oruebel, line 275:
```suggestion
   * @brief Returns the size of the dataset or attribute for each dimension.
   * @param path The location of the dataset or attribute in the file
   * @return The shape of the dataset or attribute.
```

src/nwb/NWBFile.cpp
```diff
@@ -172,6 +170,43 @@ Status NWBFile::createFileStructure(const std::string& identifierText,
...
+      NWB::Device device = NWB::Device(devicePath, m_io);
+      device.initialize("description", "unknown");
+
+      NWB::ElectrodeGroup elecGroup = NWB::ElectrodeGroup(electrodePath, m_io);
+      elecGroup.initialize("description", "unknown", device);
+    }
+  }
+
+  // write electrodes information to datasets
+  // (requires that ElectrodeGroup data is written)
```
oruebel, line 204: If the ElectrodeGroup data must be written, should this function then call ``m_io->flush()`` here first?

oruebel:
I added a couple of minor comments (mostly updates to docs), but otherwise this looks good. 

Note, there will be conflicts with #163

src/nwb/NWBFile.cpp
stephprince, line 204:
```suggestion
  // (requires that ElectrodeGroup data is initialized)
```

stephprince:
@oruebel I responded to the last comment, I think this is ready to be merged once you approve

oruebel (approved):

PR #166, opened 2025-02-24
Support reading of unknown neurodata_types with Container/Data as fallback

oruebel:
- [X] Update the RegisteredType::create methods to support fallback type for group/dataset with unknown type
- [x] `Data` and `VectorData` do not seem to appear in the registry of types, so the fallback to data fails right now. See #168  and #167

## Changes merged from #167 

This PR implements Option 1 to fix the registry for ``Data`` by creating non-templated base classes ``Data`` and then a typed variant ``DataTyped<DTYPE>``. And then the same also for ``VectorData``. With this fix, ``Data`` and ``VectorData`` are correctly registered with the registry. 

- [X] Created  non-templated base classes ``Data`` and then a typed variant ``DataTyped<DTYPE>``. This allows us to correctly register ``Data`` and only the non-templated version of ``Data`` is being registered (similar to how currently only ``Data<std::any>`` would be registered. We can still use the templated version in DEFINE_REGISTERED_FIELD definitions to fix data types for read. 
- [X] Created  non-templated base classes ``VectorData`` and then a typed variant ``VectorDataTyped<DTYPE>`` (same as for ``Data)
- [X] Added unit tests to check for all types that they are indeed registered correctly
- [X] Update the developer docs to clarify this new approach to using templates with ``RegisteredTyped``


oruebel:
This PR is also ready for review, however, we'll need merge #167 with this PR first to fix the remaining errors before merging with main.

stephprince (approved):

stephprince:
This section of the PR looks good to me, but I have not looked at #167 that should be merged with this one before merging into main.

stephprince (approved):

PR #167, opened 2025-02-24
fix data registry for Data and VectorData

### LINKED ISSUE #168: Data and VectorData are not registered correctly with the type registry
I ran into this bug as part of working on #166.  ``Data` and ``VectorData`` do currently not appear in the ``RegisteredType`` registry of types. After a few rounds of debugging,  it appears the issue is that because ``Data`` and ``VectorData`` are templated, the compiler/runtime doesn't run the ``REGISTER_TYPE_IMPL`` call that is defined in the cpp file.  Maybe this is because for templated classed can generally not be split into hpp and cpp files. Placing the  ``REGISTER_TYPE_IMPL`` into the header file does not work, as this leads to linker errors. One possible solution may be to introduce non-templated base classes for ``Data`` and ``VectorData``. We can then register those base classes. We can then created new templated subtypes of `Data` and `VectorData` to use for special case where we want to fix the data type for read at compile time.

oruebel:
Fix #168

This PR implements Option 1 to fix the registry for ``Data`` by creating non-templated base classes ``Data`` and then a typed variant ``DataTyped<DTYPE>``. And then the same also for ``VectorData``. With this fix, ``Data`` and ``VectorData`` are correctly registered with the registry. 

- [X] Created  non-templated base classes ``Data`` and then a typed variant ``DataTyped<DTYPE>``. This allows us to correctly register ``Data`` and only the non-templated version of ``Data`` is being registered (similar to how currently only ``Data<std::any>`` would be registered. We can still use the templated version in DEFINE_REGISTERED_FIELD definitions to fix data types for read. 
- [X] Created  non-templated base classes ``VectorData`` and then a typed variant ``VectorDataTyped<DTYPE>`` (same as for ``Data)
- [X] Added unit tests to check for all types that they are indeed registered correctly
- [X] Update the developer docs to clarify this new approach to using templates with ``RegisteredTyped``

**Alternative:** An alternative approach that appears to work is to simply place the call for ``template<>
REGISTER_SUBCLASS_IMPL(Data<std::any>)`` into a cpp file that we know is being run, e.g, placing the call into the ``ElementIdentifiers.cpp`` file seemed to correctly register the current templated version of ``Data<std::any>``. Similarly, it should work if we explicitly create an instance of Data<std::any>(...) somewhere else.  I opted against this approach because it: 1) feels hacky to place part of the definition of a class into an unrelated file or to create and instance that we don't need just to make sure the compiler/linker do what we need them to do, and 2) we only want to register ``Data<std::any>`` anyways with the registry and the solution implemented in this PR makes this more explicit by registering only ``Data`` but not ``DataTyped``. The main disadvantage of the approach of two classes is that it could be a bit confusing for users when to use which class, however, I don't think that is an issue, because anytime with use `VectorData` we can also choose to use `VectorDataTyped<..>`.


oruebel:
@stephprince this PR is ready for review now.

docs/pages/devdocs/registered_types.dox
```diff
@@ -236,19 +236,50 @@
...
- * and \ref AQNWB::NWB::Data "Data<DTYPE>" as these classes manage a particular dataset. Using
+ * way. AqNWB uses  templated neurodata_type classes for \ref AQNWB::NWB::VectorDataTyped "VectorDataTyped<DTYPE>" 
+ * and \ref AQNWB::NWB::DataTyped "DataTyped<DTYPE>" as these classes manage a particular dataset. Using
  * this approach, we specify the data types to use with the class directly as part of 
- * `DEFINE_REGISTERED_FIELD` so that the user doesn't need to manually specify the data type 
- * for read. The registration of templated classes follows the same main steps as above, but with a few 
- * additional considerations:
+ * \ref DEFINE_REGISTERED_FIELD macro so that the user doesn't need to manually specify the data type 
+ * for read. To implement the use of templated classes for read, we can take two main approaches
+ * by either \ref implement_templated_registered_type_two_classs or \ref implement_templated_registered_type_single_class .
```
stephprince, line 244:
```suggestion
 * by either \ref implement_templated_registered_type_two_class or \ref implement_templated_registered_type_single_class .
```

docs/pages/devdocs/registered_types.dox
```diff
@@ -236,19 +236,50 @@
...
+ * 
+ * \subsubsection implement_templated_registered_type_two_classs Using a base class and templated child class
```
stephprince, line 246:
```suggestion
 * \subsubsection implement_templated_registered_type_two_class Using a base class and templated child class
```

docs/pages/devdocs/registered_types.dox
```diff
@@ -257,7 +288,18 @@
...
+ *    \note 
+ *    In C++ the implementation of templated classes is not easily separate into ``.hpp`` and ``.cpp``
+ *    files. However, the ``template<> REGISTER_SUBCLASS_IMPL(VectorData<std::any>)`` cannot be part 
+ *    of the ``.hpp`` file where the class is being defined. Also, the compiler will only include the 
+ *    call if the ``VectorData<std::any>`` is actually being instantiated. A simple instantiation
+ *    of ``template class VectorData<std::any>;`` in the ``VectorData.cpp`` may not be sufficient 
+ *    for this. As a work-around, the ``template<> REGISTER_SUBCLASS_IMPL(VectorData<std::any>)`` may need to 
+ *    be placed in to a different ``.cpp`` file that we know is going to be built. I.e., when using
+ *    the approach of a single template class, you should confirm that the registration is being 
+ *    executed correctly by creating a unit test that checks that the type has been registered.
```
stephprince, line 302: This note/section is a little unclear to me. I thought based on our discussion and the PR summary that we did not want to place the `template<>REGISTER_SUBLCASS_IMPL(Data<std::any>)` in a different file `.cpp` file. Would it be sufficient to recommend that unit tests check a new type has been actually registered and go with the two class implementation if not?

src/nwb/file/ElectrodeTable.hpp
```diff
@@ -115,11 +115,11 @@ class ElectrodeTable : public DynamicTable
   /**
    * @brief The group names column for write
    */
-  std::unique_ptr<VectorData<std::string>> m_groupNamesDataset;
+  std::unique_ptr<VectorData> m_groupNamesDataset;
```
stephprince, line 118: I'm not sure I'm following the correct usage of `VectorData` vs. `VectorDataTyped`. Could you explain why `VectorDataTyped<std::string>` should not be used here?

src/nwb/hdmf/table/ElementIdentifiers.cpp
```diff
@@ -1,9 +1,12 @@
 #include "nwb/hdmf/table/ElementIdentifiers.hpp"
+// #include "nwb/hdmf/table/VectorData.hpp"
 
 using namespace AQNWB::NWB;
 
 // Initialize the static registered_ member to trigger registration
 REGISTER_SUBCLASS_IMPL(ElementIdentifiers)
+// REGISTER_SUBCLASS_IMPL(Data)
+// REGISTER_SUBCLASS_IMPL(VectorData)
```
stephprince, line 9:
```suggestion

using namespace AQNWB::NWB;

// Initialize the static registered_ member to trigger registration
REGISTER_SUBCLASS_IMPL(ElementIdentifiers)
```

src/nwb/file/ElectrodeTable.hpp
oruebel: Using `VectorDataTyped<std::string>` would also work here since the the data type for group name is fixed. Essentially, you want to use the templated version `VectorDataTyped<...>` anywhere you want to fix the data type for read at compile time. Currently, this is really only important when we want to read data. 

When I updated the code, I focused on making sure that `VectorDataTyped<...>` was used where we expose data for read. For code is concerned with write it didn't really make a difference whether we use `VectorData` or `VectoDataTyped`. Happy to change that to `VectorDataTyped`.

oruebel: Ok, now I remember why I left it as `VectorData` here.  The main reason is that we are using `std::unique_ptr` here. When we have a pointer of `VectorDataTyped*` and we pass it to parameter that requires `Vector*`, then C++ can cast the pointer since we `VectorDataTyped` inherits from `VectorData`. However,  `std::unique_ptr<Derived>` does not automatically convert to `std::unique_ptr<Base>`, i.e., although `VectorDataTyped` derives from `VectorData`, the `unique_ptr` template type itself is not covariant, so we wrap  a `Derived*` in a `unique_ptr<Derived>`, there is no implicit conversion to `unique_ptr<Base>`, so code that expects a `unique_ptr<VectorData>&` cannot accept a `unique_ptr<VectorDataTyped<T>>`.

So this means, while we can use `std::unique_ptr<VectorDataTyped<std::string>> m_locationsDataset;` in the class, we will then need to explictly cast it to the base `std::unique_ptr<VectorData>` when we want to call, e.g., `DynamicTable.addColumn` which expects `VectorData`:

```
auto upcasted = std::unique_ptr<VectorData>(static_cast<VectorData*>(m_locationsDataset.release()));
Status locationStatus = addColumn(upcasted, m_locationNames);
``` 

We could naturally also overload `addColumn` to define a templated version of `addColumn<DTYPE>` that would accept `VectoDataTyped<DTYPE>`, which would then cast it to `VectorData` and call the regular `addColumn`

Long story short, using `VectorDataTyped` in place of `VectorData` is fine if you know the data type to use, but once we wrap those in `std::uniqyue_ptr`, C++ no longer autmatically casts the pointers for us but we need to explicitly release the pointer and cast it.

docs/pages/devdocs/registered_types.dox
oruebel: > that we did not want to place the `template<>REGISTER_SUBLCASS_IMPL(Data<std::any>)` in a different file `.cpp` file

For AqNWB itself, I did did not want to do this because it is confusing and the two class approach works seems much cleaner. I mainly added the note to explain how it would work if someone else  would want to build on AqNWB (e.g., to create an extension) and want to use the single templated class approach to help with the weird Gotcha.

src/nwb/file/ElectrodeTable.hpp
oruebel: I added a note to clarify this to the developer docs in # 95eeaf0

docs/pages/devdocs/registered_types.dox
oruebel: I updated the docs to mention that the approach using two classes is the recommended approach. I kept the docs about the single-class approach to help avoid gotchas in the future (since this problem is not trivial to debug)

oruebel:
@stephprince I believe I have addressed your comments. Can you please take another look.

src/nwb/file/ElectrodeTable.hpp
stephprince: thanks for explaining and adding the note!

stephprince (approved):

PR #171, opened 2025-02-28
Add simple read demo

### LINKED ISSUE #170: Enhance ReadDataWrapper to expose shape information
Before actually loading any data via `ReadDataWrapper.values` it would be useful to be able to get basic information about the data object, e.g., the data type, shape, chunking etc. This will require expanding the `BaseIO` and `HDF5IO` to be able to query for storage object metadata. 

Use case:
* If we want to load only subsets of a timeseries, then it is useful to be able to first determine its shape

Updates:
* When we add this, we should also update the demo from #171 to get the shape information from the ReadDataWrapper before loading the data. Optionally, we should also update to show how to use the information to process that with arbitrary data type.
* Create #178  as a separate follow-up issue since supporting `dtype` is a bit more complex and also not as critical, since the dtype is not most cases not need until we actually load data, whereas shape is something we need to know to actually slice the data for data load.

### LINKED ISSUE #172: Enhance DataBlockGeneric to allow use of std::variant
The proposal is to make it easier for users to use `std::variant` and `std::visit`, the `DataBlockGeneric` should have a function to cast the data to an ``std::variant`` (similar to how we can cast the data to a boost::multiarray). Something along the lines of

**Type.h**
```
using NumericVariant = std::variant<
    std::vector<int8_t>, std::vector<uint8_t>,
    std::vector<int16_t>, std::vector<uint16_t>,
    std::vector<int32_t>, std::vector<uint32_t>,
    std::vector<int64_t>, std::vector<uint64_t>,
    std::vector<float>, std::vector<double>>;
```

**DataBlockGeneric add:**
```
NumericVariant asNumericVariant(const AQNWB::IO::DataBlockGeneric& dataBlock) {
    if (dataBlock.typeIndex == typeid(double)) {
        return std::any_cast<const std::vector<double>&>(dataBlock.data);
    } else if (dataBlock.typeIndex == typeid(float)) {
        return std::any_cast<const std::vector<float>&>(dataBlock.data);
    } else if (dataBlock.typeIndex == typeid(int32_t)) {
        return std::any_cast<const std::vector<int32_t>&>(dataBlock.data);
    } else if (dataBlock.typeIndex == typeid(uint32_t)) {
        return std::any_cast<const std::vector<uint32_t>&>(dataBlock.data);
    } else if (dataBlock.typeIndex == typeid(int64_t)) {
        return std::any_cast<const std::vector<int64_t>&>(dataBlock.data);
    } else if (dataBlock.typeIndex == typeid(uint64_t)) {
        return std::any_cast<const std::vector<uint64_t>&>(dataBlock.data);
    } else {
        throw std::runtime_error("Unsupported data type");
    }
}
```

This should allow users to do basic analysis with `std::visit` without having to do long if/else checks for data types, e.g.,:

```
double computeMean(const AQNWB::IO::DataBlockGeneric& dataBlock) {
    // Wrap the data in a variant
    NumericVariant dataVariant = wrapDataBlock(dataBlock);

    // Define the visitor inline and use std::visit
    return std::visit([](const auto& data) -> double {
        if (data.empty()) {
            throw std::invalid_argument("The data vector is empty");
        }
        double sum = std::accumulate(data.begin(), data.end(), 0.0);
        return sum / data.size();
    }, dataVariant);
}

int main() {
    // Example usage with HDF5IO and computeMean
    std::shared_ptr<AQNWB::IO::BaseIO> io = std::make_shared<HDF5IO>("example.h5");
    AQNWB::IO::ReadDataWrapper<AQNWB::Types::StorageObjectType::Dataset> wrapper(io, "/path/to/dataset");

    io->open();

    try {
        auto dataBlock = wrapper.valuesGeneric();
        double mean = computeMean(dataBlock);
        std::cout << "Mean: " << mean << std::endl;
    } catch (const std::exception& e) {
        std::cerr << "Error: " << e.what() << std::endl;
    }

    io->close();
    return 0;
}
```

oruebel:
- [X] Added a simple demo program for reading a file from DANDI and inspecting the ElectricalSeries in the NWB file. I defined this here as a standalone demo program (rather than as part of the tests) to allow us to show how a simple app based on AqNWB can look.  
- [X] Added test workflow to build the example and run it on one of the NWB files generated by the unit tests

As part of creating the example I identified a few features to help us simplify the example and make AqNWB easier to use for reading arbitrary data. 
- [X] To Fix #170 and avoid having to read the full data before being able to inspect the shape of the data I:
    - [X] Added `ReadDataWrapper.getShape` and `ReadDataWrapper.getNumDimensions` functions. 
    - [X] Updated the demo to use the new  `ReadDataWrapper.getShape`  to check shape before loading data
- [X] Updated the example to use `nwbfile.findOwnedFields(....)` defined on #175. This makes it easier for users to search for objects directly via the Container classes, rather than having to use the `io` object directly. 
    - [X] Updated `RegisteredType.findOwnedTypes` to allow specification of the search mode and updated the docs. This change allows us to search the full file from the `NWBFile` object, rather than just searching the types that it directly owns.
- [X] To simplify the analysis of data where the user may not know the data type until runtime (e.g., when reading data from DANDI) it is useful to be able to load the data as an `std::variant` (at least for all the common data types that NWB uses). In this way users can use `std::visit` to implement analytics (e.g., to compute a mean) in a way that is agnostic to the specific data type. 
    - [X] Added `BaseDataType::BaseDataVectorVariant` to define a `std::variant` that can represent a 1D `std::vector` of any supported `BaseDataType`
    - [X] Added `DataBlockGeneric::as_variant` method to cast data to a  `BaseDataType::BaseDataVectorVariant` that we can compute on
    - [X] Added unit tests for `DataBlock` and `DataBlockGeneric`
    - [X] Updated data read example to show an  example where we compute the mean value for an `ElectricalSeries.data` using `std::variant` and `std::visit`
    - [X] Updated the user and developer docs for read to discuss the use `std::variant`
    - [X] Updated the demo to illustrate how we can process the data using `std::visit` and `std::variant` when we don't know the data type. We here compute the global mean of the data using the `std::variant` approach and use the typed `Boost::MulitArray` to compute on the individual channels, so we can show both approaches
- [X] Added `DataBlock.getBaseDataType` and `BaseDataType.fromTypeIndex` to allow us to get the `BaseDataType` from the type id. We could technically store the `BaseDataType` also as a variable in the `DataBlock` and determine it at compile (since `DataBlock` has the `DTYPE` as a template parameter), but I figured to allow `DataBlock` to hold arbitrary types (not just `BaseDataType`) it may be useful to leave this to the runtime. This change is not strictly required for the example, but it makes `DataBlock` more consistent with `DataBlockGeneric` and the overall API. 

Fix #170
Fix #172

Notes: This PR depends on #175, i.e., we should either merge this PR with #175 or merge #175 first and then this PR.


demo/inspect_electrical_series/README.md
```diff
@@ -0,0 +1,110 @@
+# NWB Analysis Demo
+
+This is a simple C++ demo project that shows how to read and analyze NWB (Neurodata Without Borders) files using the aqnwb library. The demo reads the file `sub-EF0147_ses-20190204T144339_behavior+ecephys.nwb` and performs basic statistical analysis on the electrophysiology data.
```
oruebel, line 3:
```suggestion
This is a simple C++ demo project that shows how to read and analyze NWB (Neurodata Without Borders) files using the aqnwb library. The demo reads a NWB file and performs basic statistical analysis on the ElectricalSeries electrophysiology data.
```

demo/inspect_electrical_series/README.md
```diff
@@ -0,0 +1,109 @@
...
+- Boost library
+- aqnwb library (built from the parent project)
+
+## Download an example NWB dataset that contains ElectricalSeries data
+
+E.g. https://api.dandiarchive.org/api/assets/4c440a73-9250-45bb-a342-a7da4d01b2fd/download?content_disposition=inline to
+download a single file from DANDIset 000232 at https://dandiarchive.org/dandiset/000232/0.240510.2038
+
+Alternatively, we can also run AqNWB test suite and use a suitable file from the tests, e.g,. 
+`build/dev/tests/data/createElectricalSeries.nwb`
```
stephprince, line 19:
```suggestion
## Identify an example NWB dataset that contains ElectricalSeries data

1. For this example we will download a single file from [dandiset 000232](https://dandiarchive.org/dandiset/000232/0.240510.2038) using this [download link](https://api.dandiarchive.org/api/assets/4c440a73-9250-45bb-a342-a7da4d01b2fd/download?content_disposition=inline). https://api.dandiarchive.org/api/assets/4c440a73-9250-45bb-a342-a7da4d01b2fd/download?content_disposition=inline to
download a single file from DANDIset 000232 at https://dandiarchive.org/dandiset/000232/0.240510.2038
2. Place the file in the `aqnwb/demo/inspect_electrical_series` directory.
```

I think for the example readme it would be helpful to keep consistent using the DANDI file throughout.

demo/inspect_electrical_series/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Building the Demo
+
+1. Make sure you have built the main aqnwb project first 
+   following the instructions in the AqNWB documentation
+
+2. Create a build directory for the demo:
+
+```bash
+# From the  demo/inspect_electrical_series directory
```
stephprince, line 29:
```suggestion
cd demo/inspect_electrical_series
```

demo/inspect_electrical_series/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+```bash
+# Basic configuration
+cmake ..
+
+# Or with custom paths to dependencies
+cmake .. \
+  -DAQNWB_DIR=../../build/dev \
+  -DHDF5_DIR=../../libs/hdf5_build \
+  -DBOOST_ROOT=../../libs/boost_1_82_0
```
stephprince, line 44:
```suggestion
  -DAQNWB_DIR=../../../build/dev \
  -DHDF5_ROOT=../../libs/hdf5_build \
  -DBOOST_ROOT=../../libs/boost_1_82_0
```

You could update the path flag to HDF5_ROOT to keep consistent with the environment variable name used in the main installation instructions

demo/inspect_electrical_series/README.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+## Running the Demo
+
+After building, you can run the demo:
+
+```bash
+# From the demo/DANDISET_000232/build directory
+cd bin
+./nwb_analysis_demo ../../sub-EF0147_ses-20190204T144339_behavior+ecephys.nwb
```
stephprince, line 57: When using the dandi demo file, I get this error:

```
Analyzing ElectricalSeries at: /processing/ecephys/LFP/ElectricalSeries_97
Number of time points: 7409
Number of channels: 64
Data unit: volts
Data description: no description
HDF5-DIAG: Error detected in HDF5 (1.14.3) thread 0:
  #000: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5D.c line 1061 in H5Dread(): can't synchronously read data
    major: Dataset
    minor: Read failed
  #001: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5D.c line 1008 in H5D__read_api_common(): can't read data
    major: Dataset
    minor: Read failed
  #002: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5VLcallback.c line 2092 in H5VL_dataset_read_direct(): dataset read failed
    major: Virtual Object Layer
    minor: Read failed
  #003: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5VLcallback.c line 2048 in H5VL__dataset_read(): dataset read failed
    major: Virtual Object Layer
    minor: Read failed
  #004: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5VLnative_dataset.c line 373 in H5VL__native_dataset_read(): can't read data
    major: Dataset
    minor: Read failed
  #005: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5Dio.c line 401 in H5D__read(): can't read data
    major: Dataset
    minor: Read failed
  #006: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5Dchunk.c line 2919 in H5D__chunk_read(): unable to read raw data chunk
    major: Low-level I/O
    minor: Read failed
  #007: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5Dchunk.c line 4548 in H5D__chunk_lock(): data pipeline read failed
    major: Dataset
    minor: Filter operation failed
  #008: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5Z.c line 1355 in H5Z_pipeline(): required filter 'deflate' is not registered
    major: Data filters
    minor: Read failed
  #009: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5PLint.c line 267 in H5PL_load(): can't find plugin. Check either HDF5_VOL_CONNECTOR, HDF5_PLUGIN_PATH, default location, or path set by H5PLxxx functions
    major: Plugin for dynamically loaded library
    minor: Object not found
  #010: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5PLpath.c line 804 in H5PL__find_plugin_in_path_table(): search in path /usr/local/hdf5/lib/plugin encountered an error
    major: Plugin for dynamically loaded library
    minor: Can't get value
  #011: /Users/smprince/Documents/projects/hdf5/hdf5-1.14.3/src/H5PLpath.c line 857 in H5PL__find_plugin_in_path(): can't open directory (/usr/local/hdf5/lib/plugin). Please verify its existence
    major: Plugin for dynamically loaded library
    minor: Can't open directory or file
libc++abi: terminating due to uncaught exception of type H5::DataSetIException
zsh: abort      ./nwb_analysis_demo ../../sub-EF0147_ses-20190204T144339_behavior+ecephys.nwb
```

It seems some filter plugins may be required to run this demo?

demo/inspect_electrical_series/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 3.15)
+project(nwb_analysis_demo VERSION 0.1.0 LANGUAGES CXX)
+
+# Set C++ standard
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Allow configuring paths to dependencies
+set(AQNWB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../build/dev" CACHE PATH "Path to aqnwb build directory")
+set(HDF5_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../libs/hdf5_build/install/cmake" CACHE PATH "Path to HDF5 build directory")
```
stephprince, line 10:
```suggestion
set(HDF5_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../libs/hdf5_build/install/cmake" CACHE PATH "Path to HDF5 build directory")
```

I think HDF5_ROOT is the environment variable used by CMake in FindHDF5. This is also what we tell users to set in the main aqnwb installation instructions

stephprince (commented):
Looks good! Added my review

src/nwb/NWBFile.cpp
```diff
@@ -39,8 +39,11 @@ NWBFile::NWBFile(std::shared_ptr<IO::BaseIO> io)
 NWBFile::NWBFile(const std::string& path, std::shared_ptr<IO::BaseIO> io)
     : Container("/", io)  // Always use "/" for the path
 {
-  std::cerr << "NWBFile object is always the root. Path must be /" << std::endl;
-  assert(path == "/");
+  if (path != "/") {
+    std::cerr << "NWBFile object is always the root. Path must be /"
+              << std::endl;
+    assert(path == "/");
+  }
```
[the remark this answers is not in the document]
oruebel: > I'm guessing this is to trigger a program error?

Correct. 

> This logic is a little confusing to me since it checks `path != "/"` and then asserts that `path == "/"`

The reason for the difference is due to how `assert` and `if statements work. `assert` is designed to trigger when a condition is False whereas `if` is designed to trigger when a condition is True. Using just the `assert` with a message makes sense. I think I had mainly left the logic this way to separate logging from assert, but this is fine.

.github/workflows/build-demo.yml
```diff
@@ -0,0 +1,64 @@
...
+    # could have the workflow depend on the unit test workflow and
+    # download its artifacts, but its simple enough to just run the tests
+    # - name: Download test data artifact
+    #   uses: actions/download-artifact@v4
+    #   with:
+    #     name: test-files-macos-latest
+    #     path: build/tests/data  
+    - name: Test
+      working-directory: build
+      run: ctest --output-on-failure --no-tests=error -C Release -j 2
```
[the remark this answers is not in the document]
oruebel: In terms of capabilities, the unit tests should cover things. However, since the demos are building small apps there is the caveat that they have their own cmake build configurations, so it could make sense to run them on all platforms. 

I'm only using a single file generated by the unit tests. I.e., if we want to run these tests across platforms it may be easiest to just make a separate script to generate an NWB file that we can use to test the `nwb_analysis_demo` script. This script wouldn't actually need to be with AqNWB either, but could be a simple Python script with PyNWB to create an example file with some ElectricalSeries in it. This would require some additional code, but would keep the test logic simpler and avoid dependencies between the unit tests and the tests for the demos. 

For now, I think this is fine, and I'll open an issue so we can address this separately.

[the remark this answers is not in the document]
oruebel: I filed https://github.com/NeurodataWithoutBorders/aqnwb/issues/186 for this

demo/inspect_electrical_series/README.md
oruebel: Interesting. It worked for me. Not sure why this is failing.

demo/inspect_electrical_series/README.md
oruebel: Done

demo/inspect_electrical_series/CMakeLists.txt
oruebel: Done

oruebel: I reverted this back to HDF5_DIR because otherwise the GitHub action was failing. I think the differnce is that `HDF5_DIR` points to the cmake folder and `HDF5_ROOT` points to the main HDF5 folder such that cmake searches for the build. By setting HDF5_ROOT in the CMakeLists.txt it confuses the workflow, but by using HDF5_DIR here we can set it to a custom path and cmake can still find HDF5 based on HDF5_ROOT if HDF5 is in a different location. The alternative would be remove the set calls foir HDF5_DIR and BOOST_DIR here  and always require the user to set them.

oruebel:
@stephprince I've committed your suggestions and synced the branch with `main`.

demo/inspect_electrical_series/README.md
oruebel: The `deflate` filter that the message mentions is `gzip`, which should be installed by default with HDF5.

stephprince: I was using a version of HDF5 that was built from source, which doesn't seem to include the filter plugins by default. The other filter tests that you had added were also failing for similar reasons.

Updated to use the brew installation version instead and everything works!

demo/inspect_electrical_series/CMakeLists.txt
stephprince: I see. I don't think it's super important to change in this PR since this is just a demo, but I would lean towards having the user set them if needed (especially since a specific version of boost is currently specified as the default). In the workflows, the BOOST_ROOT set in the CMakeLists.txt file is currently being ignored for compatibility reasons and I think if we remove the default path for HDF5_DIR in the CmakeLists.txt file, then Cmake should be able to find HDF5 (as in the test workflows)?

PR #175, opened 2025-03-03
Simply search for neurodata_type objects via RegisteredType

### LINKED ISSUE #173: Allow search for any type via BaseIO::findTypes
Currently ``BaseIO::findTypes`` does not support searching for any neurodata_type, only to search for a specific list of known neurodata_types. A user should be able to provide an empty list of types to search for objects with an arbitrary neurodata_type. This is useful when we want to discover all objects with a type

### LINKED ISSUE #174: Allow searching for typed objects owned by a type
It would be useful to be able to find for a given `RegisteredType` object, all the child objects that have a neurodata_type that it owns, i.e,. find all the children of a type that have the current object as its first parent. For  this we should:
1. Enhance `BaseIO::findTypes` to add an option `exclude_starting_path` to allow us to skip the starting path while still STOP_AT_TYPE stops searching once it finds a type. Currently if the starting path is itself a typed object,  STOP_AT_TYPE will always just return the starting path. 
2. Add a function to `RegisteredType` to call `getIO()->findTypes(getPath(), types=any type, search_mode=STOP_ON_TYPE, exclude_starting_path=true)
3. This also required #173

oruebel:
Fix #173
Fix #174 

- [X] Updated `BaseIO.findTypes` to add a `bool exlude_starting_path` option to allow us to search for all objects within a type but not include the ``starting_path`` itself. This also prevents `STOP_ON_TYPE` from stopping at the `starting_path`
- [X] Updated `BaseIO.findTypes` to allow `types={}` as input to indicate that we want to find any object with a `neurodata_type` independent of what the specific type is.
- [X] Added `RegisteredType.findOwnedTypes` to allow us to find all neurodata_type objects that are owned by the current object, i.e., these are the children with a neurodata_type that have this object as their closest parent with a neurodata_type
- [X] Updated unit tests

### DIFF
--- a/tests/testNWBFile.cpp
+++ b/tests/testNWBFile.cpp
@@ -55,8 +55,13 @@ TEST_CASE("initialize", "[nwb]")
   initStatus = nwbfile.initialize(generateUuid());
   REQUIRE(initStatus == Status::Success);
   REQUIRE(nwbfile.isInitialized());
+
+  // Since we didn't create any typed objects within the NWBFile, we should
+  // have no owned types
+  auto result = nwbfile.findOwnedRegisteredTypes();
+  REQUIRE(result.size() == 0);
+
   nwbfile.finalize();
-  io->close();
 }
 
 TEST_CASE("createElectricalSeries", "[nwb]")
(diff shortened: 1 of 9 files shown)


oruebel:
@stephprince this PR is also ready for review

tests/testNWBFile.cpp
```diff
@@ -346,6 +362,7 @@ TEST_CASE("createMultipleEcephysDatasets", "[nwb]")
   }
 
   nwbfile.finalize();
+  io->close();
```
stephprince, line 365:
```suggestion
```

`nwbfile.finalize();` should be closing the io object so I think these additions are not needed

tests/testNWBFile.cpp
```diff
@@ -401,6 +418,7 @@ TEST_CASE("createAnnotationSeries", "[nwb]")
...
```
stephprince, line 421:
```suggestion
```

tests/testNWBFile.cpp
```diff
@@ -441,4 +459,5 @@ TEST_CASE("setCanModifyObjectsMode", "[nwb]")
 
   // stop recording
   nwbfile.finalize();
+  io->close();
```
stephprince, line 462:
```suggestion
```

stephprince (approved):
Looks good to me! Very minor suggestions about some extra lines that I got added to the tests but are unrelated to the changes in this PR

stephprince (approved):

PR #177, opened 2025-03-16
Support use of std::variant and std::visit

### LINKED ISSUE #172: Enhance DataBlockGeneric to allow use of std::variant

oruebel:
To simplify analysis of data where the user may not know the data type until runtime (e.g., when reading data from DANDI) it would be nice to be able to load the data as an std::variant (at least for all the common data types that NWB uses). In this way users can use std::visit to implement analytics (e.g., to compute a mean) in a way that is agnostic to the specific data type. 

- [X] Added `BaseDataType::BaseDataVectorVariant` to define a `std::variant` that can represent a 1D `std::vector` of any supported `BaseDataType`
- [X] Added `DataBlockGeneric::as_variant` method to cast data to a  `BaseDataType::BaseDataVectorVariant` that we can compute on
- [X] Added unit tests for `DataBlock` and `DataBlockGeneric`
- [X] Updated data read example to show a  example where we compute the mean value for an `ElectricalSeries.data` using `std::variant` and `std::visit`
- [X] Updated the user and developer docs for read to discuss the use `std::variant`

Fix #172

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ CMakeUserPresets.json
 compile_commands.json
 CMakeCache.txt
 CMakeFiles/
+
+# locally install libraries 
+libs/
--- a/docs/pages/devdocs/read_design.dox
+++ b/docs/pages/devdocs/read_design.dox
@@ -52,14 +52,19 @@
  *     \ref AQNWB::IO::DataBlock "DataBlock" is templated on the specific data type, and
  *     stores the \ref AQNWB::IO::DataBlock::data "data" as an appropriately typed 1-dimensional 
  *     ``std::vector`` along with the \ref AQNWB::IO::DataBlock::shape "shape" of the data.
- * 6. ``BOOST::multi_array``
+ * 6. [BOOST::multi_array](https://www.boost.org/doc/libs/1_83_0/libs/multi_array/doc/user.html)
  *   -  ``BOOST::multi_array`` can also be used simplify access to multi-dimensional data.
  *      The \ref AQNWB::IO::DataBlock::as_multi_array "DataBlock.as_multi_array"
  *      convenience method generates a ``boost::const_multi_array_ref<DTYPE, NDIMS>`` for us.
+ * 7. [std::variant](https://en.cppreference.com/w/cpp/utility/variant)
+ *   - ``std::variant` can also be used when we want to compute on the data in a type-safe
+ *     manner but do not know the data type beforehand (e.g., when reading NWB data from a
+ *     third party).
  *
  * \note
- * The \ref AQNWB::IO::DataBlock::fromGeneric "DataBlock.fromGeneric" and
- * \ref AQNWB::IO::DataBlock::as_multi_array "DataBlock.as_multi_array" methods use casting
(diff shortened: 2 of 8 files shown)


oruebel:
@stephprince this PR is ready for review now

oruebel:
@stephprince I merge this PR with #171 because I need this to finish up that PR

PR #181, opened 2025-03-18
Create Namespace registry

### LINKED ISSUE #169: Create docs on how to implment extension
With the current design of AqNWB, I believe we are already pretty close to enabling other to use AqNWB to implement C++ classes for reading/writing of neurodata_types defined in extensions. To provide basic support for enabling read/write of extensions I believe we need the ability to:

- [X] Create custom classes for extension types. With the `RegisteredType` facility and [developer docs on implementing neurodata_types](https://neurodatawithoutborders.github.io/aqnwb/registered_type_page.html) we already have a solid foundation for this. 
- [X] Read data even if we don't have a custom class implemented (see #166) 
- [ ] Save extension schema in an NWB file. This seems doable but may require a bit of additional code. We have the [`generate_spec_files.py`](https://github.com/NeurodataWithoutBorders/aqnwb/blob/main/resources/generate_spec_files.py) utility, and maybe it can already consume extension schema (or be updated to do so. Once we can generate the C++ files for the schema, all we would need is a way for the user to add them to the NWB file, so probably just a convenience method on `NWBFile`
- [ ]  Document for users/developers the steps needed for implementing an extension in AqNWB. This can probably be brief since we can point to the existing docs on `RegisteredType` and the steps to create and save extension schema can hopefully be short.
     - **Optional** Create an example of how to implement a very simple extension (e.g., an extension that extends `Device` to add a new field)

### LINKED ISSUE #179: Support management of namespace schema
To ease integration of extensions we need to make sure that extension schema are reliably cached in the file. To do this, it would be useful to define a static registry for namespaces, such that all namespaces that we use are managed.

### LINKED ISSUE #180: Generalize generate_spec_files.py to ease use for extensions
To simplify the implementation of extensions it would be nice to generalize the `generate_spec_files.py` script to make it easy to generate schema header files for arbitrary extension schema.

oruebel:
This PR is to simplify and clarify the process for how to implement an NWB extension in NWB. In particular this PR revamps how we manage schema/namespaces to simplify the process of integrating new schema. 

## Issues resolved by this PR

Fix #169
Fix #179 
Fix #180

## Tasks
- [X] Created a new `NamespaceRegistry`. The registry is populated statically by the compiler and collects all namespaces used. This allows us to simplify the integration of additional namespaces from extensions and makes the management of namespaces more transparent overall.
- [X] Updated the existing spec header files to use call `REGISTER_NAMESPACE` to register the main namespaces
- [X] Updated the existing spec files to use a ``const std::vector`` instead of a ``constexpr std::array`` to simplify use of different namespaces with different numbers of spec files
- [X] Updated `NWBFile::cacheSpecifications` and `NWBFile::Initialize` to use the new `NamespaceRegistry` directly 
- [X] Refactored the `generate_spec_files.py` script to make it easier to use for others as a command line tool, improve documentation, and simplify configuration for use with extensions 
- [X] Updated  `generate_spec_files.py` to support the new `NamespaceRegistry`
- [X] Add developer docs on how to use the `NamespaceRegistry`
- [X] Updated REGISTERED_TYPE macro to support setting the namespace name via the `const std::string namespaceName` variable from the spec header files

## Optional task (to be addressed in separate PR)
- [X] Create a demo for writing an extension. This has been implemented in PR #183 (issue #182)

### DIFF
--- a/docs/pages/devdocs/install.dox
+++ b/docs/pages/devdocs/install.dox
@@ -40,8 +40,23 @@
  *
  * \note
  * If you are using custom installations of **HDF5** or **BOOST** that are not being detected
- * automatically by cmake, you can specify `HDF5_ROOT` and `BOOST_ROOT` environment variables to
- * point to install directories of HDF5 and BOOST respectively.
+ * automatically by CMake, you should set the `CMAKE_PREFIX_PATH` variable to include the install
+ * directories of HDF5 and BOOST. This can be done either as an environment variable or directly
+ * on the CMake command line. For example:
+ *
+ * \code{.sh}
+ * # Option 1: Set as environment variable (colon-separated on Unix/macOS)
+ * export CMAKE_PREFIX_PATH=/path/to/hdf5_install:/path/to/boost_install
+ * cmake --preset=dev
+ *
+ * # Option 2: Pass as a CMake variable (semicolon-separated)
+ * cmake --preset=dev -DCMAKE_PREFIX_PATH=/path/to/hdf5_install;/path/to/boost_install
+ * \endcode
+ *
+ * This ensures CMake can find the necessary config files for both HDF5 and Boost.
+ * 
+ * The use of `HDF5_ROOT` and `BOOST_ROOT` environment variables is deprecated for modern CMake
+ * and may not work reliably with recent CMake versions.
  *
  * \section  devbuild_presets_subsec Developer Presets
  *
@@ -114,4 +129,3 @@
  * - `spell-fix` :  run the `codespell` tool on the codebase with `FIX=YES` to both check and automatically fix  common spelling errors
  * - `docs` : builds the documentation using Doxygen. (Note: run `cmake --preset=dev -DBUILD_DOCS=ON` before building to add docs target)
  */
-


oruebel:

resources/generate_spec_files.py
```diff
@@ -1,101 +1,182 @@
...
+        except Exception as e:
+            logger.error(f"Failed to load YAML file {filepath}: {e}")
+            raise
+    elif filepath.suffix == '.json':
+        try:
+            with open(filepath) as f:
+                return json.load(f)
+        except Exception as e:
+            logger.error(f"Failed to load JSON file {filepath}: {e}")
+            raise
```
stephprince, line 37:
```suggestion
```

Since this script now generates the header files directly from the schema .yaml files, I think we no longer need to include the json files or support loading them.

resources/generate_spec_files.py
```diff
@@ -1,101 +1,182 @@
...
-                    chunk_var_names = []
-                    for i, chunk in enumerate(chunks):
-                        chunk_var_name = f'{var_name}_part{i}'
-                        with open(header_file, 'a') as fo:
-                            fo.write(f'constexpr std::string_view {chunk_var_name} = R"delimiter({chunk})delimiter";\n')
-                        chunk_var_names.append(chunk_var_name)
+                schema_file = namespace_file.parent / s['source']
+                # If the schem file is missing, check if it was converted to JSON/YAML
+                if not os.path.exists(schema_file):
+                    for ext in ['.yaml', '.yml', '.json']:
```
stephprince, line 151:
```suggestion
                    for ext in ['.yaml', '.yml']:
```

See previous comment about supporting the .json files

resources/generate_spec_files.py
```diff
@@ -1,101 +1,182 @@
...
+    Process schema files in the specified directory and generate corresponding C++ header files.
 
+    Args:
+        schema_dir (Path): Directory containing the schema files.
+        output_dir (Path): Directory to output the generated header files.
+        chunk_size (int): Size of the chunks for splitting large JSON strings.
+    """
+    logger.info(f"Starting to process schema files in directory: {schema_dir}")
+    for file in schema_dir.rglob(r"*namespace.*"):
+        if file.suffix in ['.yaml', '.yml', '.json']:
```
stephprince, line 171:
```suggestion
        if file.suffix in ['.yaml', '.yml']:
```

See comment above about supporting the json files. I was running into errors when pointing to the .json directory anyway.

stephprince (commented):
Looks good! 

Most of my comments are related to `generate_spec_files.py` - I think we should update the files in the `src/spec/` folder to be generated by the most recent version of this script. When I regenerated the core and hdmf-common header files, I currently get some validation errors. I think my comments should address most of those but should double check after these updates.

These updates should also allow us to remove the `resources/spec` folder entirely and work directly with the yaml files from nwb-schema.

resources/generate_spec_files.py
```diff
@@ -1,101 +1,182 @@
...
+from typing import List, Dict, Union
+import os
+
+# Setup logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+def load_file(filepath: Path) -> Union[Dict, List]:
+    """
+    Load a YAML or JSON file and return its content.
```
oruebel, line 15:
```suggestion
    Load a YAML schema file and return its content.
```

oruebel:
@stephprince thanks for the helpful suggestion. I have now addressed the remaining issues with `generate_spec_files.py` that you had identified. Please take a look.

oruebel:
> These updates should also allow us to remove the `resources/spec` folder entirely and work directly with the yaml files from nwb-schema.

Good idea. I've implemented this change in https://github.com/NeurodataWithoutBorders/aqnwb/pull/199 I created a separate PR because I needed to branch of another PR in order to fix the workflows and pick up all the other required changes to python scripts.

PR #185, opened 2025-03-19
Generate classes

### LINKED ISSUE #184: Create script to generate AqNWB RegisteredType classes directly from schema
Similar to how [generate_spec_files.py](https://github.com/NeurodataWithoutBorders/aqnwb/blob/main/resources/generate_spec_files.py) generates header files for the schema, it would be nice to also have a script to generate `RegisteredType` classes for all the neurodata_types in the schema. At the very least, one should be able to create skeletons for the classes to make it easier to fill out the classes. In principle, one should be able to also create the `DEFINE_FIELD` read definitions as well as large parts of the `initialize` methods.

oruebel:
Fix #184 

## Purpose: 
This PR adds a script to auto-generate templates for AqNWB classes directly from NWB schema. The intent of this script is to help developers create new classes by providing at least an outline of a class that can be edited and fixed. The code that the script is generating is by no means perfect, but it should provide helpful skeletons that one can fix up and the code can also provide a better starting point for AI coding tools.

## Example:
To generate skeleton classes for all neurodata_types of the NWB schema we can run:

```
python resources/schematype_to_aqnwb.py resources/spec/core/2.7.0/nwb.namespace.json  test_output
```

or for a simpler example, we can run it on the labmetadata_extension_demo
 
```
python resources/schematype_to_aqnwb.py demo/labmetadata_extension_demo/spec test_output
```

## Comment:
While it is in principle possible to fully auto-generate the code (similar to how MatNWB does it), this would require much more significant effort to make sure that all the various corner cases are covered. Also, for C++ this will be a bit trickier since it is a statically typed language (whereas Python and MatLab are dynamically typed). Also, I don't have the time to implement a fully-automated version. However, I think the current script will at least help simplify the process of integrating new types.


oruebel:

oruebel:
**NOTE** if #190 is being merged first, then we need to update this PR to use the new DEFINE_ATTRIBUTE_FIELD and DEFINE_DATASET_FIELD macros instead of DEFINE_FIELD

oruebel:
**NOTE** I updated this PR for the class generation to be in compliance with #190 which changed the DEFINE_FIELD_MACRO by splitting it into a DEFINE_ATTRIBUTE_FIELD and DEFINE_DATASET_FIELD macro. I.e., this PR should ideally be merged after #190. However the changes only affect the class generation script so as long as #190 gets merged not too long after this, then its also Ok to merge this PR first.

PR #207, opened 2025-08-19
Test schema upgrade

### LINKED ISSUE #205: Add workflow for testing spec generation
The GitHub workflows proposed in https://github.com/NeurodataWithoutBorders/aqnwb/pull/199/files get the latest schema version and test our python utility functions run successfully. Ideally we could extend these to check that the spec header files can be included to generate a valid minimal NWBfile.

oruebel:
Fix #205 

Add workflow to test that the NWB and HDMF schema can be upgraded to the latest release.


oruebel:
@stephprince this PR is now also ready for review

stephprince:
> @stephprince should we remove "- name: Run generate_nwb_schema_headers.sh" from thepython-utils.yml workflow here?

I agree would be fine to remove that step from the `python-utils.yml` workflow now since this workflow tests that script. Pushed a commit that does so

PR #212, opened 2025-08-25
DRAFT: Explore multiton for registered types

oruebel:
This PR is to evaluate whether `RegisteredType::create` should look up if an object with the same path and I/O already exists by looking up the object via `io->getRecordingObjects()->getRecordingObject(path)` and return the existing object. 

➕  This would avoid creating duplicate objects to prevent creation of different objects that modify the same file object.
➖ This sort of `multiton` pattern may be too restrictive for users. 
➖  In addition to `create` this will also require some rethinking of the templated RegisteredType class, e.g., `VectorDataTyped`. For these we may have an untyped `VectorData` first for which we then create `VectorDataTyped` for read only. This would not be allowed by  `RegisteredType::create`. However, is this needed? If yes, should `VectorDataTyped` be declared a read-only type to avoid issues?

I think there is a balance to be found here between preventing the user of the API from potentially doing bad things by creating multiple objects for the same path/io and limiting flexibility of the API. I am currently leaning towards documenting the best practice and not restricting the API in this way. @stephprince thoughts?

### DIFF
--- a/src/nwb/RegisteredType.cpp
+++ b/src/nwb/RegisteredType.cpp
@@ -72,12 +72,25 @@ std::string RegisteredType::getNamespace() const
   return "";
 }
 
+std::shared_ptr<RegisteredType> RegisteredType::getExistingInstance(const std::string& path, std::shared_ptr<AQNWB::IO::BaseIO> io)
+{
+  auto recordingObjects = io->getRecordingObjects();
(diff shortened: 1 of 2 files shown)


stephprince:
We agreed we will document the best practice for now and can consider this in the future.

PR #214, opened 2025-08-26
Update ElectrodesTable type

oruebel:
- [X] Rename `ElectrodeTable` to `ElectrodesTable`
- [X] Update `ElectrodesTable` to not use `DynamicTable` as typename (i.e., use `REGISTER_SUBCLASS` instead of `REGISTER_SUBCLASS_WITH_TYPENAME`


docs/pages/devdocs/registered_types.dox
```diff
@@ -286,10 +286,10 @@
  * 2. A class for a modified type that does not have its  own `neurodata_type` 
  *    in the NWB schema. An example is `ElectrodesTable` in NWB <v2.7, which
```
stephprince, line 287:
```suggestion
 *    in the NWB schema. An example is `ElectrodesTable` in NWB <v2.9, which
```

stephprince (commented):
I think this may impact read of NWB files with an earlier version of the schema <2.9. However since we had previously discussed that we will only support read of files produced by AqNWB and we have not had an official release this is fine, but is something we may want to consider in the future.

oruebel:
> I think this may impact read of NWB files with an earlier version of the schema <2.9. However since we had previously discussed that we will only support read of files produced by AqNWB and we have not had an official release this is fine, but is something we may want to consider in the future.

@stephprince I created https://github.com/NeurodataWithoutBorders/aqnwb/issues/215 to discuss this. I think the impact on read for older files should be manageable and I think could be addressed with minimal extra logic.

PR #216, opened 2025-08-27
Read  ElectrodesTable for NWB <=2.8

### LINKED ISSUE #215: Support `ElectrodesTable` read for NWB files `<=2.8`
#214 updated to support the change in NWB v.2.9, which add a typename `ElectrodestTable`. It would be useful to test that reading `<v2.8` `ElectordeTable` worksI think it affects read in the following way:

1. If a user explicitly uses [`ElectrodeTable::create(io)` ](https://github.com/NeurodataWithoutBorders/aqnwb/blob/9b8d9b63dc95a44ba67cc5c661d151f2329801ed/src/nwb/file/ElectrodeTable.hpp#L44-L48) (this in on #209) or calls [`RegisteredType::create("core::ElectricalSeries,  path, io) `](https://github.com/NeurodataWithoutBorders/aqnwb/blob/9b8d9b63dc95a44ba67cc5c661d151f2329801ed/src/nwb/RegisteredType.cpp#L75-L79) to read the `ElectrodeTable`, then I think that will work fine, because in those cases the user explicitly specifies the class they want to use for reading, and since the structure of the `ElectrodeTable` has not changed, the read should work fine. 
2. If the users uses `RegisteredType::create(path, io)` where AqNWB discovers the type from the the file, then for older files that would mean that the function would create a `DynamicTable` instance instead of the more specific `ElectrodeTable`. Read should still work, but `DynamicTable` would not have the specific details (e.g,. the `readLocationColumn`) methods. 

To make sure that the user always gets an `ElectrodeTable` instance, we could update this logic:

https://github.com/NeurodataWithoutBorders/aqnwb/blob/9b8d9b63dc95a44ba67cc5c661d151f2329801ed/src/nwb/RegisteredType.cpp#L123-L130

to add an explicit check that would look something like this:

```
if(path == ElectrodeTable::electrodeTablePath && fullClassName == "core::DynamicTable"){
    fullClassName = "core::ElectrodesTable"
}
```

oruebel:
Fix #215 

- [X] Moved code for reading the type of an object to `BaseIO::getFullTypeName` to make it reusable
- [X] Updated `BaseIO::getFullTypeName` to map to `ElectrodesTable` for NWB <=2.8 files
- [X] Updated `RegisteredType::create` to use  `BaseIO::getFullTypeName`
- [X] Renamed `ElectrodesTable::electrodeTablePath` to `electrodesTablePath` for consistency
- [X] Added unit test for the new behaviors

### DIFF
--- a/src/io/BaseIO.cpp
+++ b/src/io/BaseIO.cpp
@@ -62,30 +62,24 @@ Status BaseIO::createCommonNWBAttributes(const std::string& path,
   return Status::Success;
 }
 
-std::string BaseIO::getFullTypeNameFromFile(const std::string& path)
+std::string BaseIO::getFullTypeName(const std::string& path)
 {
-  std::cerr << "DEBUG BaseIO::getFullTypeNameFromFile path=" << path
-            << std::endl;
   // Read the "namespace" attribute
   AQNWB::IO::DataBlockGeneric namespaceData =
       readAttribute(AQNWB::mergePaths(path, "namespace"));
   auto namespaceBlock =
       AQNWB::IO::DataBlock<std::string>::fromGeneric(namespaceData);
   std::string typeNamespace = namespaceBlock.data[0];
-  std::cerr << "DEBUG BaseIO::getFullTypeNameFromFile typeNamespace="
-            << typeNamespace << std::endl;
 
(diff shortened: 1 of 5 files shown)


oruebel:

stephprince (approved):

PR #220, opened 2025-09-03
add CHANGELOG for 0.1.0 release

stephprince:
add CHANGELOG describing features, bug fixes, and other changes during development so far (generated via Claude from git commit history)


oruebel (approved):
Looks good. It would be nice to also include this in the *For Developer* section of the docs, similar to how we include the legal docs in https://github.com/NeurodataWithoutBorders/aqnwb/blob/main/docs/pages/devdocs/legal.dox . We can fix this here or make a separate issue add that later.

oruebel:
> It would be nice to also include this in the _For Developer_ section of the docs

I just added a commit to add the Changelog to the docs

PR #224, opened 2025-09-23
Enhance `schematype_to_aqnwb` to create proper file hierarchy

oruebel:
- [X] Updated `schematype_to_aqnwb` to place the generated source files into a subfolder structure based on the namespace and schema file (see example below), instead of all source files being placed into the same folder.
- [X] Updated `schematype_to_aqnwb.py` to used `pathlib` to improve handling of paths
- [X] Updated the github workflow for testing `schematype_to_aqnwb.py`  accordingly
- [x] Updated the changelog

**Example file hiearchy``
```
output_dir/
├── core/                    # Core namespace folder
│   ├── base/               # From nwb.base.yaml
│   │   ├── TimeSeries.hpp
│   │   ├── TimeSeries.cpp
│   │   └── ...
...
└── hdmf-common/            # HDMF-common namespace folder (included by the nwb namespace)
    ├── base/               # From hdmf-common base schemas
    │   ├── Data.hpp
    │   ├── Data.cpp
    │   └── ...
    └── table/              # From hdmf-common table schemas
        ├── DynamicTable.hpp
        ├── DynamicTable.cpp
        └── ...
```


resources/utils/schematype_to_aqnwb.py
```diff
@@ -491,18 +518,43 @@ def parse_schema_file(file_path: str) -> Tuple[SpecNamespace, Dict[str, Spec]]:
...
         if hasattr(spec, "neurodata_type_def") and spec.neurodata_type_def is not None:
             neurodata_types[spec.neurodata_type_def] = spec
+            
+            # Get the source file for this spec using the catalog method
+            source_file = namespace.catalog.get_spec_source_file(spec_name)
+            if source_file:
+                type_to_file_map[spec.neurodata_type_def] = Path(source_file)
+            else:
+                # Fall back to namespace file if no specific source file found
+                type_to_file_map[spec.neurodata_type_def] = namespace_path
```
stephprince, line 541: IIUC I think the default here should be none if there is no source file? Otherwise would this potentially create a "namespace" subfolder in the hierarchy

resources/utils/schematype_to_aqnwb.py
```diff
@@ -491,18 +518,43 @@ def parse_schema_file(file_path: str) -> Tuple[SpecNamespace, Dict[str, Spec]]:
...
+            
+            # Look through all namespaces in the catalog to find the original namespace
+            for ns_name in type_map.namespace_catalog.namespaces:
+                ns = type_map.namespace_catalog.get_namespace(ns_name)
+                if ns != namespace and spec_name in ns.catalog.get_registered_types():
+                    # This type is originally from a different namespace
+                    type_namespace = ns_name
+                    break
+            
+            type_to_namespace_map[spec.neurodata_type_def] = type_namespace
```
stephprince, line 555: A `get_namespace_for_type` method will be added to the namespace catalog in https://github.com/hdmf-dev/hdmf/pull/1312/files#diff-0458c07608cd941f7b8f1cfa52c52faa016b6c2a17e913e820a2a5108609776dR336. I think it behaves similarly to this section, and could potentially replace some of this code

oruebel: Thanks for the helpful suggestion. Since this has not been released in HDMF yet, I created a new issue https://github.com/NeurodataWithoutBorders/aqnwb/issues/226 to update this code once the release is available.

stephprince (approved):

PR #227, opened 2025-09-30
Pip install for tools

oruebel:
- [X] Added `pyproject.toml` for python utils to make them pip installable
- [X] Add `uv` header to the python scripts to make them usable via `uv`
- [X] Added new `aqnwb_utils.py` utility to provide a single command-line utility for aqnwb
- [X] Updated the developer docs to clarify usage of the python utilities
- [X] Updated the github workflows to use pip to install the command line utilities
- [X] Updated `schematype_to_aqnwb.py` to create a more robust CMake file for the example app, because locating aqnwb relative to the script location doesn't work if the utilities are installed somewhere else using pip.
- [X] Updated the changelog


stephprince:
I am wondering if this would be a good application of using uv to declare script dependencies and make the python file an executable. That way we could avoid creating an additional python package files and these scripts would be relatively self-contained. 

I think the main tradeoff would be if we do plan to add several python utilities then it would be good to use more traditional packaging methods

oruebel:
@stephprince thanks for merging #229 with this PR. I've made a few more updates here to make sure the docs are consistently using uv for the python utilities and to avoid duplication between `resources/utils/README.md` and the doxygen docs. This PR is now ready for review.

stephprince (approved):

PR #229, opened 2025-10-01
Uv install tools

oruebel:
Update #227 to use UV


oruebel:
@stephprince sorry, I didn't see #228 until just now. This PR now updates the approach from #227 to support uv. It still also supports pip by also having a pyproject.toml (instead of the setup.py). Let me know what you think

stephprince:
I was thinking uv would allow us to keep it minimal without the additional `pyproject.toml` and `aqnwb_utils.py` files, but I am good either way. I agree if we want it to be pip installable we should use `pyproject.toml` instead of `setup.py`

oruebel:
Up to you if you want to keep pip or not. I just figured folks may not be familiar with uv, but these are developer tools so I think it's fine. Just let me know if I should remove the pip part

PR #233, opened 2025-10-16
Fix windows warnings

### LINKED ISSUE #234: Warnings in Windows build workflow
Currently the build on Windows in the github Action shows a large number of warnings, e.g., https://github.com/NeurodataWithoutBorders/aqnwb/actions/runs/18537764983/job/52838135034?pr=232 (see the Build step)

oruebel:
Fix #234

**Changes:**

- [X] Resolved various compiler warnings on Windows (`-Wmaybe-uninitialized`, `-Wsign-conversion`, `-Wshadow`, `-Wdeprecated-copy`, `-Wcatch-value`) and fixed cross-platform build issues related to `std::filesystem` linkage.
- [X] Fixed error in `test_ecephys_data_read` where the wrong version of `compute_mean` was being called, leading to a missed lines in codecov and ultimately resulting in a misleading example
- [X] Update CHANGELOG

**Note:** The Build tab in https://github.com/NeurodataWithoutBorders/aqnwb/actions/runs/18572523516/job/52958228537?pr=233 shows that most of the warnings are resolved, which makes the output cleaner and easier to parse. There are still a few remaining warnings, however, these are outside of the scope of AqNWB:

- __`-Wold-style-cast` warnings:__ These are coming from macros within the HDF5 library's header files (e.g., `H5T_VARIABLE`). Even though we use `static_cast` in our code, the warning is triggered by the old-style cast inside the HDF5 macro itself. We cannot fix this without modifying the HDF5 library.
- __`-Wmaybe-uninitialized` warnings:__ These warnings originate from the C++ standard library's implementation of `std::regex`. This seems to be a false positive or an issue within the standard library code, not in `aq-nwb`.

### DIFF
--- a/src/io/hdf5/HDF5IO.cpp
+++ b/src/io/hdf5/HDF5IO.cpp
@@ -1148,7 +1148,7 @@ std::vector<SizeType> HDF5IO::getStorageObjectShape(const std::string path)
   try {
     H5::DataSet dataset = m_file->openDataSet(path);
     dataspace = dataset.getSpace();
-  } catch (H5::Exception& e) {
+  } catch (H5::Exception&) {
     // Read the attribute
     std::unique_ptr<H5::Attribute> attributePtr = this->getAttribute(path);
     dataspace = attributePtr->getSpace();


stephprince:
I think the order of operations specified in the target PR is to merge this into add_nwb_basetypes first and then merge the combined PR into main

Issue #247, opened 2026-01-23
Removing boost as a dependency


chittti:
Hi-- thanks for your work on this project!

We're using AqNWB, and in an effort to reduce the list of dependencies, refactored it to remove boost as a dependency. Instead, used the C++ standard library with minimal additional code. 

The primary concern with this refactor was `boost/multi_array`. The replacement code is concise, only because a small part of it was being used, specfically `boost::const_multi_array_ref`. So a new class, `ConstMultiArrayView` was added to `ReadIO.hpp`, which seems to work well. Rest of the new code is in `Utils.hpp`.

If there were plans to use a lot more of boost's facilities that haven't been used yet, then I understand this effort isn't desirable. You can take a look at our fork [here](https://github.com/BlackrockNeurotech/aqnwb) and passing tests [here](https://github.com/BlackrockNeurotech/aqnwb/actions). 

If this looks ok, I can make a PR.

Thanks!

PR #248, opened 2026-01-23
Modify CMake so transitive includes from hdf5::hdf5_cpp and Boost::boost propagate correctly

### LINKED ISSUE #246: Make headers public for publicly linked boost and HDF5
Currently, HDF5 and Boost are both linked publicly. This means that the symbols from those libraries are available to downstream applications that link aqnwb. However, because the libraries are linked via `${HDF5_CXX_LIBRARIES}` and `${Boost_LIBRARIES}`, this excludes the headers.

The HDF5 and boost headers are made available to aqnwb via `include_directories(${HDF5_INCLUDE_DIRS})` and `include_directories(SYSTEM ${Boost_INCLUDE_DIRS})`, but this does not re-export them in aqnwb's includes. Downstream targets can't actually leverage those symbols unless they independently find the headers. It's a bit of an anti-pattern to ask aqnwb-consumers who also want HDF5 to `find_package(HDF5)` just for the headers, and specifically DO NOT link the library too.

There are a couple options here to make this more friendly for downstream apps:

1. Try using `Boost::boost` instead of `${Boost_LIBRARIES}`, and `hdf5::hdf5` instead of `${HDF5_CXX_LIBRARIES}`.
    * I'm pretty sure these targets bring their headers with them (Boost:boost is header-only).
    * `hdf5::hdf5` requires CMake 3.19, which might not come for free on your CI runner
2. Use `target_include_directories` with `PUBLIC` instead of `include_directories`.
3. Fully hide the symbols so downstream apps must manage boost/hdf5 on their own if they need it.
    * I think this might make sense for boost; I've seen having public boost conflict with other libraries (e.g. Matlab).
    * I think this makes less sense for hdf5 because you want to make sure the app's understanding of an hdf5 container matches aqnwb's.

Let me know what the preferred solution is and if you want me to make an attempt at it.

cboulay:
Modify CMake so transitive includes from hdf5::hdf5_cpp and Boost::boost propagate correctly
Fixes #246 

Now, downstream targets that do

```cmake
find_package(aqnwb)
target_link_libraries(my_target PRIVATE aqnwb::aqnwb)
find_dependency(HDF5 COMPONENTS C CXX)
```

will have access to boost and hdf5 symbols **and headers**.

IMO, boost should be private, but I couldn't do that here because boost is used in the aqnwb headers.
Whether this is important might become moot depending on how #247 is dealt with.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Removed Boost as a dependency  (@chittti, [#250](https://github.com/NeurodataWithoutBorders/aqnwb/pull/250) )
 * Updated `DataBlock::as_multi_array()` to return `ConstMultiArrayView` instead of `boost::const_multi_array_ref` to remove the need Boost and for C++17/20 compatibility  (@chittti, [#250](https://github.com/NeurodataWithoutBorders/aqnwb/pull/250))
 * Updated the docs and examples to discuss the optional use of `std::mdspan` if C++23 is used (@oruebel, [#250]((https://github.com/NeurodataWithoutBorders/aqnwb/pull/250))
+* Changed the dependency includes in CMake so the HDF5 C++ headers and libraries found via `${HDF5_INCLUDE_DIRS}` / `${HDF5_CXX_LIBRARIES}` are exported to AqNWB consumers (@cboulay, [#248](https://github.com/NeurodataWithoutBorders/aqnwb/pull/248)).
 
 ### Fixed
 * ...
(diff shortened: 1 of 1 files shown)


cboulay:
CHANGELOG entry added. Of course I'll have to rebase and modify everything after boost is removed.

oruebel:
> I rebased and squashed to a single commit.

Thank you @cboulay I'll take a look.

demo/labmetadata_extension_demo/README.md
```diff
@@ -86,8 +84,8 @@ Before building the demo, ensure you have the following dependencies installed:
 
 If you encounter build errors:
 
-1. **Check dependency paths**: Ensure that the paths to AqNWB and HDF5 are correctly set in the CMake command or in the CMakeLists.txt file.
+1. **Check CMAKE_PREFIX_PATH**: Ensure that `-DCMAKE_PREFIX_PATH` points to the directory where aqnwb was installed (the directory containing `lib/cmake/aqnwb/`).
 
 2. **Check compiler compatibility**: Make sure your compiler supports C++17 features.
 
-3. **Library linking issues**: If you encounter linking errors, verify that the AqNWB library has been built successfully and that the path to it is correct.
+3. **Library linking issues**: If you encounter linking errors, verify that the aqnwb library has been built and installed successfully.
```
oruebel:
```suggestion
3. **Library linking issues**: If you encounter linking errors, verify that the aqnwb library has been built and installed successfully.
 
4. **HDF5 not found**: While the HDF5 dependency is transitive, CMake does not search non-standard locations for dependencies unless you tell it where to look. I.e., if HDF5 is installed in a non-standard location, you will also need to include the HDF5 prefix path, e.g, `cmake .. -DCMAKE_PREFIX_PATH="/path/to/aqnwb/install;/path/to/hdf5/install/cmake"`. 
```

Same suggestion as above, just for the README of the other demo

PR #300, opened 2026-07-02
Update schema to NWB 2.10

### LINKED ISSUE #284: Add support for new MeaningsTable type in hdfm-common
### Overview

In https://github.com/hdmf-dev/hdmf-common-schema/pull/91 the following key changes were made:

1) added `MeaningsTable` as a new type 
2) updated`DynamicTable` to add a sub-group for storing instances of `MeaningsTable` 

With this design, each column in a `DynamicTable` may optionally have a corresponding `MeaningsTable`. 

### Approach

- [ ] Update schema in AqNWB to support the latest `hdmf-common` schema. See http://nwb.org/aqnwb/nwb_schema_page.html for instructions. 
- [ ] Add new class for representing `MeaningTable`. See http://nwb.org/aqnwb/registered_type_page.html for instructions. 
- [ ] Update `DynamicTable::initialize` implement the required schema changes, in particular, creating the required `meanings_tables` group 
- [ ] Update `DynamicTable` to add the required macros (e.g., DEFINE_UNNAMED_REGISTERED_FIELD) for reading and creating `MeaningsTable` objects
- [ ] Optional: Add convenience methods in `DynamicTable` to simplify creation of `MeaningsTables` for a column
- [ ] Update the user-docs to add a tutorial on using `DynamicTable` and `MeaningsTable`, incl., adding an example in tests/examples
- [ ] Add comprehensive unit tests for the new features
- [ ] Update the CHANGELOG

### LINKED ISSUE #285: Add support for EventsTable
### Overview

In https://github.com/NeurodataWithoutBorders/nwb-schema/pull/645 the following key changes are made:

1) added new types `EventsTable`, `TimestampVectorData`, ` DurationVectorData`
2) deprecated `AnnotationSeries`

### Approach

- [ ] Update schema in AqNWB to support the latest `core` NWB schema. See http://nwb.org/aqnwb/nwb_schema_page.html for instructions. 
- [ ] Add new classes for representing `EventsTable`, `TimestampVectorData`, ` DurationVectorData`. See http://nwb.org/aqnwb/registered_type_page.html for instructions. 
- [ ] Add methods to `EventsTable` to simplify recording of events since this will require update to multiple columns in a table
- [ ] Update `NWBFile` to add the required macros (e.g., DEFINE_UNNAMED_REGISTERED_FIELD) for reading and creating `EventsTable` objects
- [ ] Optional: Add convenience methods in `NWBFile` to simplify creation of `EventsTable` for acquisition
- [ ] Update the user-docs to add a tutorial on using `EventsTable` for recording events`, incl., adding an example in tests/examples
- [ ] Add comprehensive unit tests for the new features
- [ ] Update the CHANGELOG

### LINKED ISSUE #298: Update to NWB 2.10 schema
Update to the latest NWB 2.10 schema

### LINKED ISSUE #299: Add support for new TimestampVectorData and DurationVectorData
TimestampVectorData and DurationVectorData are new VectorData types required for implementation of EventsTable

### LINKED ISSUE #303: Support row-wise recording of DynamicTable
In the current model, if we want to record rows of a `DynamicTable`, the user needs to manually update all columns of the table to add a set of rows. For the metadata-centric tables, e.g., ElectrodesTable this is Ok, since we can usually create the table once at the beginning. However for the new `EventsTable` types we will likely need to record data continuously to the table as new events are being generated. For convenience, it would be nice if a user could provide all data for a set of rows (across all columns) and then have the write be handled, rather than having to manually update columns. This should also update the ElementIdentifiers.

### LINKED ISSUE #306: Update MeaningsTable to store target as refererence attribute
https://github.com/hdmf-dev/hdmf/pull/1525
https://github.com/hdmf-dev/hdmf-common-schema/pull/97

Change the schema for MeaningsTable to store the `target` not as a link but as an object reference attribute. This will require updating the schema and update the MeaningsTable.

oruebel:
Fix #298 
Fix #306

Update schema version to the latest releases
- [X] Updated `spec/core.hpp` to NWB schema 2.10
- [X] Updated `spec/hdmf_common.hpp` to schema version ~~1.9~~ 1.10
    - [X] `MeaningsTable.target` is being changed to an object reference attribute in hdmf-common 1.10 and should be changed here accordingly before merge. See https://github.com/hdmf-dev/hdmf-common-schema/pull/97. Fix via [79b6fbf](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/79b6fbf72b4d0fb63b9ad9a515bd96b62ffbc1b3) here and also updated on https://github.com/NeurodataWithoutBorders/aqnwb/pull/304 to fix the actual `MeaningsTable` class.
- [X] Updated `spec/hdmf_experimental.hpp` to schema version 0.6

## From #301 : Add event VectorData types

Fix #299 

- **Update `schematype_to_aqnwb.py` to facilitate generation the new types**
    - [X] Fixed indent of REGISTER_SUBCLASS call and use the proper core and hdmf-common namespace when used as included schema
    - [X] Updated minimum Python version for `resources/utils` to ensure use of the latest versions of PyNWB with uv
    - [X] Updated cpp code generation to use new std::weak_ptr via `getIO` instead of the previous `m_io` shared pointer
- **Implement the new  `TimestampVectorData` and  `DurationVectorData`  types:**
    - [X] Implemented `nwb/event/TimestampVectorData.hpp/cpp` 
    - [X]  Implemented `nwb/event/DurationVectorData.hpp/cpp` 
    - [X] Added unit tests for `TimestampVectorData`
    - [X] Added unit tests for `DurationVectorData`
- **Update and test that `DynamicTable` properly supports the new column types:**
    - [X]  Updated `DynamicTable.readColumn` using type traits to to support reading of both: i) VectorData columns with a specific data type and ii) columns that are a subtype of VectorData (e.g.,  `TimestampVectorData` and  `DurationVectorData`)
    - [X] Added override for `DyanmicTable.addColumn` to allow adding of fully configured VectorData columns and derived column types
   - [X] Added unit test to ensure `TimestampVectorData` and `DurationVectorData` can be added via`DynamicTable.addColumn` and read back via `DynamicTable.readColumn`
- **Other changes:**
    - [X] Updated the CHANGELOG  

## From #302 : Add MeaningsTable

Fix #284 

**Implement new `MeaningsTable` type:**

- [X] Implement the new `src/nwb/hdmf/table/MeaningsTable.*` source for `MeaningsTable`
- [x] Implement comprehensive unit tests for `MeaningsTable`

**Update `DynamicTable` to support adding `MeaningsTables`:**

- [X] Added "DEFINE_UNNAMED_REGISTERED_FIELD(" to create `readMeaningsTable` for reading `MeaningsTables` associated with a `DynamicTable`. This also creates `createMeaningsTableInstance` methods for creatin a new `MeaningsTables` that is not yet initialized. 
- [X] Added `DynamicTable.createMeaningsTable` methods for creating a fully initialized `MeaningsTable` that is associated with a particular column of the `DynamicTable`
- [X] Add unit tests to test the new `DynamicTable.createMeaningsTable` and `DynamicTable.readMeaningsTable` methods.
~~- [ ] Update the docs/tutorials as necessary to describe the use of MeaningsTable alongside DynamicTable~~ *This will be part of follow-up once the full stack for EventsTable has been added*

**Other Changes**

- [X] Updated `ElectrodesTable.initialize` to add a new optional parameter `rowChunkSize`.  Similar to the `MeaningsTable` the existing `ElectrodesTable` also creates pre-defined columns in `initialize`. Currently the chunking was set in `ElectrodesTable` to `1`, which is too small.  Adding  `rowChunkSize` increases the default to 100 and allows the caller to set the chunk size. 
- [X] Updated the CHANGELOG

## From #304 : Add EventsTable type

Fix #285 

**Add support for `EventsTable`**
- [X] Added new `EventTable` type
- [X] Added unit tests for the new `EventTable` type
- [X] Updated NWBFile to create `events/` group and added `NWBFile::createEventsTable` convenience function

**Enhance `DynamicTable` to improve support for use during acquisition**
- [X]  Refactored `DynamicTable` so it creates and initializes its `ElementIdentifiers` `id` column internally during `initialize()`. This is to ensure the `ElementIdentifiers` are always create before we start the recording. Calling `DynamicTable::setRowIDs()` previously required creation of `ElementIdentifiers` but may be called later during recording. As part of this change, `setRowIDs()` now takes only the row ID values and writes them directly to the table's built-in `id` column instead of accepting a separate `ElementIdentifiers` object. In practice this aligns the API with the intended usage, since callers should have always used the table's own `id` dataset.  (@oruebel, [#302](https://github.com/NeurodataWithoutBorders/aqnwb/pull/304))
- [X] Updated `DynamicTable` column-name management to be immediate and table-driven. `addColumnName()` now flushes new column names to the file as columns are added, `setColNames()` is now restricted to reordering the existing set of columns (it no longer acts as a general overwrite mechanism), and `finalize()` no longer writes the `colnames` attribute because column-name updates are persisted when they occur. This ensures that the `colnames` attribute is always consistent with the actual columns in the table and helps avoid creation of invalid files. (@oruebel, [#304](https://github.com/NeurodataWithoutBorders/aqnwb/pull/304))

**Other**
- [X] Fix bugs in `TimestampVectorData` and `DurationVectorData`; e.g, bad namespace and inconsistent ordering of parameters in initialize
- [x] Update Changelog 

## From #305 : Row-based Recordings

Fix #303 

**Problem:** The current implementation of `DynamicTable` is mainly designed for the purpose of metadata tables, e.g., `ElectrodeTable`, where we can usually fill in the whole table before the actual acquisition. With the new `EventTable`, however, we need enhanced support for acquiring the data in the table as part of the acquisition workflow. 

**Goal:** Update `DynamicTable` and `EventTable` to simplify row-based recording and configuration of the table

**Changes:**
- **Support configuration of the columns of a DynamicTable directly via initialize, rather than just incremental via`addColumn`**
    - [X]  Added `Data::DataSpecBase` and `Data::DataSpec<T>` to support runtime configuration of `DynamicTable` column layouts and to decouple table schema definition from object initialization. Each `Data` subclass (e.g., `VectorData`, `ElementIdentifiers`, `TimestampVectorData`, `DurationVectorData`) now exposes a nested `DataSpec` struct that bundles the dataset configuration, description, and any type-specific parameters needed to create and initialize the column. 
    - [X] Added `DynamicTable::createDefaultDataSpecs` static factory to return the default ordered list of `DataSpec` objects for a table (currently just the `id` column). Subclasses (`ElectrodesTable`, `EventsTable`, `MeaningsTable`) override this to append their own required columns, allowing callers to retrieve, inspect, and customize the default column configuration before passing it to `initialize()`.
    - [X] Updated `DynamicTable::initialize`, `ElectrodesTable::initialize`, `EventsTable::initialize`, and `MeaningsTable::initialize` to accept a `std::vector<DataSpecPtr>` column-spec list instead of individual `rowChunkSize` / resolution / flag parameters.
    - [X] Added `DynamicTable::addColumn(DataSpecPtr)` overload consistent with the new `DataSpec`-based API used by `initialize()` and `createDefaultDataSpecs()`.
    - [X] Added `DyanmicTable::validateDataSpecs` (and `DynamicTable::checkRequiredColumnNames` helper function) to validate column specifications as part of `DyanmicTable::initialize` before initialization. Also updated subclassed of `DynamicTable` (e.g., `ElectrodesTable`, `EventsTable`, `MeaningsTable`) to override `validateDataSpecs` to provide their specific validation logic. 
- **Support row-based write of DynamicTable across all columns of the table**
    - [X] Added `DynamicTable::addRow` and `DynamicTable::addRows` methods to support row-based data insertion into a `DynamicTable`. Rows are specified as `DynamicTable::RowData` (an `unordered_map` from column name to `CellValue` variant), enabling type-safe, column-keyed writes without requiring callers to manage per-column buffers directly. Row IDs are auto-generated if not provided. 
    - [X] Added `BaseDataType::BaseDataVariant` scalar variant type and `BaseDataType::createEmptyVectorVariant` helper to support runtime-typed single-cell and buffer operations needed by `addRow`/`addRows`. (@cline, @oruebel, [#305](https://github.com/NeurodataWithoutBorders/aqnwb/pull/305))
- **Update Documentation** 
    - [X] Added tutorial to demonstrate acquisition workflow for DynamicTable. The tutorial uses EventsTable as an example and shows: 1) row-based appending of data and 2)column-based write and addition of full columns to a table, and 3) adding of a `MeaningsTable` to a columns 
    - [X] Updated the CHANGELOG
- **Created Follow-up issues** 
    - [X] Appending rows for reference columns is not yet working. See https://github.com/NeurodataWithoutBorders/aqnwb/issues/313
    - [X] Update `schema_to_aqnwb.py` utility to follow the new structure for `DynamicTable`. See https://github.com/NeurodataWithoutBorders/aqnwb/issues/314


oruebel:
@rly note the changes to the files in "src/spec" are auto-generated via scripts so you can mostly ignore those.

resources/utils/schematype_to_aqnwb.py
```diff
@@ -93,8 +93,15 @@ def render_define_registered_field(
...
         std::string parentPath = AQNWB::mergePaths(m_path, "{field_path}");
         std::string objectPath = AQNWB::mergePaths(parentPath, objectName);
-        if (m_io->objectExists(objectPath)) {{
-        return std::make_shared<{neurodata_type}>(objectPath, m_io);
+        auto ioPtr = getIO();
+        if (ioPtr == nullptr) {
+            std::cerr << "IO object has been deleted. Can't read {neurodata_type}: "
+                      << m_path << std::endl;
+            return nullptr;
+        }
```
rly:
```suggestion
        if (ioPtr == nullptr) {{
            std::cerr << "IO object has been deleted. Can't read {neurodata_type}: "
                      << m_path << std::endl;
            return nullptr;
        }}
```
Need to escape braces in this template

src/nwb/event/DurationVectorData.hpp
```diff
@@ -0,0 +1,105 @@
...
+   */
+  DurationVectorData(const std::string& path,
+                     std::shared_ptr<AQNWB::IO::BaseIO> io);
+
+  /**
+   * @brief Virtual destructor.
+   */
+  virtual ~DurationVectorData() override {}
+
+  // TODO: Update the initialize method as appropriate.
```
rly:
```suggestion
```

src/nwb/NWBFile.hpp
```diff
@@ -127,11 +130,66 @@ class NWBFile : public NWBContainer
...
+   * column in seconds.
+   * @param createAnnotationColumn Whether to create the annotation column.
+   * @param rowChunkSize The chunk size for the rows of the table.
+   * @return The generated EventsTable or nullptr if failed.
+   */
+  std::shared_ptr<EventsTable> createEventsTable(
+      const std::string& name,
+      const std::string& description,
+      const std::string& sourceDescription = "",
+      float timestampResolution = 0.001f,
```
rly: I don't think we should set a default timestamp resolution of 1 ms. Users are likely to not catch this and will write inaccurate data unintentionally.

rly: Since both timestamp resolution and duration resolution are optional in the schema, can they be made optional here when initializing TimestampVectorData and DurationVectorData? Do we need to provide default values? -1 isn't bad, but I don't see a need to write it.

src/nwb/NWBFile.cpp
```diff
@@ -120,7 +120,8 @@ bool NWBFile::isInitialized() const
       "processing",
       "stimulus",
       "general",
-      "specifications"};
+      "specifications",
+      "events"};
```
rly: The events group is not required in the schema. Claude says that reading older files without this group will raise an error. `isInitialized()` returns false, and `NWBFile::initialize` then re-runs `createFileStructure` on an already-populated file

src/nwb/NWBFile.cpp
```diff
@@ -169,6 +170,7 @@ Status NWBFile::createFileStructure(const std::string& identifierText,
   ioPtr->createGroup(NWBFile::GENERAL_PATH);
   ioPtr->createGroup(mergePaths(NWBFile::GENERAL_PATH, "/devices"));
   ioPtr->createGroup(mergePaths(NWBFile::GENERAL_PATH, "/extracellular_ephys"));
+  ioPtr->createGroup(NWBFile::EVENTS_PATH);
```
rly: Note: We don't need to create this if there are no events. Either way is fine.

rly:
Do you want to update to NWB schema 2.11 here?

src/nwb/NWBFile.cpp
oruebel: That's a good point. Fixing this will require additional logic to conditionally create the `events/` group when it is needed. The same issue also applies to the `intervals/` group. I'll fix this in #325 so that i can consistently fix it for both.

src/nwb/NWBFile.cpp
oruebel: Fixed in [5ad93fe](https://github.com/NeurodataWithoutBorders/aqnwb/pull/325/commits/5ad93fe07c440fbbea2e57dd29f2a81111e4d554) in #325 which makes both `events/` and `intervals/` groups optional and creates them if necessary in the `EventTable::initialize` and `TimeIntervals::initialize` methods.

src/nwb/NWBFile.cpp
oruebel: Fixed in [5ad93fe](https://github.com/NeurodataWithoutBorders/aqnwb/pull/325/commits/5ad93fe07c440fbbea2e57dd29f2a81111e4d554) in #325 which makes both `events/` and `intervals/` groups optional and creates them if necessary in the `EventTable::initialize` and `TimeIntervals::initialize` methods.

src/nwb/NWBFile.hpp
oruebel: Makes sense. We can use `std::optional` with default `std::nullopt` to make the parameter optional and indicate that it is missing by default. Ultimately this will require changes in the code of the underlying types `EventsTable`, `TimestampVectorData`, and `DurationVectorData` since this is where these parameters are actually being processed.

oruebel: I'll also make `source_description` optional then, since it is also optional in the schema

oruebel: Another issue is that creation of `duration` column is currently tied to whether `durationResolution` is provided. I.e., currently we can only create a `duration` column if we set the `durationResolution`. This should be separated, i.e., a user should be able to create the  optional `duration` column but omit the optional `durationResolution"

oruebel: Fixed in [6f3a3b6](https://github.com/NeurodataWithoutBorders/aqnwb/pull/325/commits/6f3a3b6c088b95a6ce90fd51962d39e6f2c308e6) . I accidentally also added this in the downstream PR #325 rather than here.

oruebel:
> Do you want to update to NWB schema 2.11 here?

Sounds good. Done in [b810516](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/b810516f0f4707c66f7fb6b99c71c65d2b03a879)

src/nwb/event/EventsTable.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  std::vector<DataSpecPtr> specs =
+      DynamicTable::createDefaultDataSpecs(rowChunkSize);
+
+  IO::ArrayDataSetConfig timestampConfig(
+      IO::BaseDataType::F32, SizeArray {0}, SizeArray {rowChunkSize});
+  specs.push_back(std::make_shared<TimestampVectorData::DataSpec>(
+      "timestamp",
+      timestampConfig,
+      "A 1-D VectorData that stores timestamps in seconds from the session "
```
rly: `createDefaultDataSpecs` writes the `TimestampVectorData` **type** doc as this column's `description`, but the schema gives the `timestamp` dataset its own [column-level doc](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/2.11.0/core/nwb.event.yaml#L77-L80): "Column containing the time that each event occurred, in seconds, from the session start time."

`readTimestampColumn` uses that column-level text already (`EventsTable.hpp:118`) and same for `MeaningsTable` in both places, so as written the same column is described one way in the file and a different way through the accessor. Same for `duration`. And `annotation` gets an invented sentence where the schema says "Column containing user annotations about events."

I think it would be worth settling on one convention. I'd suggest the column-level docs here, so the `description` attribute matches the accessor and `MeaningsTable`.

rly:
I made some minor spelling fixes and updated the generated text with doc changes in schema 2.11. I will continue the review later!

src/nwb/event/EventsTable.cpp
oruebel: Good catch . Fixed in [b4405d4](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/b4405d424b93d89713757efffb598baf9af84035)

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -158,6 +407,273 @@ Status DynamicTable::finalize()
...
+      ConfiguredColumn confCol;
+      confCol.name = colName;
+      confCol.dataType = col->readData()->getDataType();
+      confCol.column = col;
+      m_configuredColumns.push_back(confCol);
+      m_configuredColumnIndices[colName] = m_configuredColumns.size() - 1;
+    }
+  }
+  return Status::Success;
+}
```
rly, line 899: If you open an existing file and call `addRows` on a table without calling `initialize()` first, the column data gets written but the call returns `Failure` and no ids are written, leaving the table ragged.

The sequence:

1. The constructor reads `colnames` into `m_colNames`, but leaves `m_rowElementIdentifiers` as `nullptr`.
2. `addRows` calls `ensureConfiguredColumnsLoaded()`, which sees `m_configuredColumns` empty and calls `loadConfiguredColumnsFromFile()`.
3. That restores the data columns by looping over `m_colNames` — but `m_colNames` never contains `"id"`, because `configureDataObject` returns at line 562 before it reaches `addColumnName`. So `m_rowElementIdentifiers` stays null.
4. `writeColumnBuffer` writes every data column to disk.
5. `setRowIDs` then hits the null check at line 256, prints "ElementIdentifiers dataset is not initialized", and returns `Failure`.

So `loadConfiguredColumnsFromFile` restores the data columns but not the `id` column. Since this function exists only to support the "table loaded from file" path, I think it should restore both:

```cpp
  m_rowElementIdentifiers = readIdColumn();
```

`recordData()` opens the dataset lazily through `getDataSet()`, so this works without `initialize()` having run, same as the data columns already do here. It also makes `generateRowIDs` pick up the existing row count and continue numbering instead of restarting at 0.

The "test row-wise append" section only reads from the re-opened table, so this path isn't covered. Worth extending it to append a row after reopening and check that the ids continue.

src/nwb/event/DurationVectorData.hpp
```diff
@@ -0,0 +1,103 @@
...
+  virtual ~DurationVectorData() override {}
+
+  /**
+   * @brief Initialize the object
+   * @param data The configuration for the dataset
+   * @param description Description of the dataset
+   * @param resolution The temporal resolution of the timestamps - in seconds.
+   * This is typically the sampling period (1 / sampling_rate), also known as
+   * the clock period, of the data acquisition system from which the timestamps
+   * were recorded or derived.
```
rly:
```suggestion
   * @param resolution The temporal resolution of the durations - in seconds.
   * This is typically the sampling period (1 / sampling_rate), also known as
   * the clock period, of the data acquisition system from which the durations
   * were recorded or derived.
```

src/nwb/hdmf/table/DynamicTable.hpp
```diff
@@ -160,26 +272,170 @@ class DynamicTable : public Container
...
+   */
+  std::vector<std::string> m_colNames;
 
   /**
    * @brief The row ids data object for write
    */
   std::shared_ptr<ElementIdentifiers> m_rowElementIdentifiers;
+
+  std::vector<ConfiguredColumn> m_configuredColumns;
+  std::unordered_map<std::string, SizeType> m_configuredColumnIndices;
```
rly, line 593: These four members hold the table's column state, and five paths mutate them, each updating a different subset. That leads to four related problems:

- `addConfiguredColumn` (`.cpp:596`) skips `m_configuredColumnIndices`, so `getConfiguredColumn` returns `nullptr` for anything added via `addColumn`/`addReferenceColumn`.
- `loadConfiguredColumnsFromFile` (`.cpp:608`) rebuilds from `m_colNames`, which never contains `"id"`, so `m_rowElementIdentifiers` stays null and `addRows` on a re-opened table fails at `setRowIDs` (separate comment).
- The `m_colNames.size() > m_configuredColumns.size()` guards (`.cpp:227/247/389`) infer membership from a size comparison, which on a re-opened table registers only the new column and leaves later `addRows` writing a ragged table.
- `configureDataObjects` (`.cpp:526`) clears the registry but not `m_colNames` or `m_rowElementIdentifiers`, so a second `initialize()` leaves `colnames` out of step with the registry.

Rather than patch each, I'd enforce the invariant in one place:

```cpp
  /**
   * @brief Register a column in the table's column registry.
   *
   * Adds the column to the ordered list of configured columns, records its
   * name-to-index lookup, and adds its name to the colnames attribute. If a
   * column of the same name is already registered, returns the existing index
   * and modifies no state.
   */
  SizeType registerColumn(const std::shared_ptr<VectorData>& column);
```

`configureDataObject`, both `addColumn` overloads, `addReferenceColumn`, and `loadConfiguredColumnsFromFile` all route through it, and the three size heuristics go away since "already present" becomes its job. Roughly 40 lines, no public API change.

Two things don't fall out of it: `loadConfiguredColumnsFromFile` still needs to restore `m_rowElementIdentifiers` via `readIdColumn()`, since `id` is excluded from `colnames`; and `configureDataObjects` should reset all four members, not two.

src/nwb/event/EventsTable.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @param durationResolution The temporal resolution of the optional duration
+   * column in seconds. See `DurationVectorData::initialize()` for more details.
+   * If a negative value is provided (default), then the duration column will
+   * not be created.
+   * @param createAnnotationColumn Whether to create the annotation column
+   * (default: false)
+   * @param rowChunkSize The chunk size for the rows of the table.
+   * @return A vector of DataSpecPtr containing the default specs.
+   */
+  /**
```
rly: These two doc blocks are adjacent with no declaration between them, so Doxygen drops the first one and `createDefaultDataSpecs` (line 71) ends up undocumented. Moving that block down to sit directly above its own declaration fixes it.

src/nwb/NWBFile.hpp
```diff
@@ -242,6 +300,12 @@ class NWBFile : public NWBContainer
                                   Get a ProcessingModule stored in the
                                       processing group)
 
+  DEFINE_UNNAMED_REGISTERED_FIELD(readEventsTable,
+                                  createEventsTable,
+                                  EventsTable,
+                                  "events",
+                                  Get an EventsTable stored in the events group)
```
rly, line 447: `NWBFile` ends up with three `createEventsTable` functions, and which one runs depends on how many arguments you pass:

- **1 argument** → the one this macro generates. It only constructs the C++ object; `RegisteredType::create` writes nothing to the file, so you get an empty handle with no group, attributes, or columns.
- **2 or more** → the hand-written overload at line 158, which creates and initializes the table for real.

Both build the same path (`/events/<name>`), so they're two entry points to the same object in different states. Both forms are in use: `testNWBFile.cpp:93` takes the initializing one, `:149` takes the handle and calls `initialize` on it afterwards.

The risk is that adding or removing an argument silently switches which function runs. `createProcessingModule` just above uses the same macro but has no hand-written overload competing with it, so `EventsTable` is the only ambiguous one here.

`DynamicTable` already avoids this: its equivalent accessor is named `createMeaningsTableInstance` (`DynamicTable.hpp:304`) so it can't collide with `createMeaningsTable`. Renaming the macro's write name here to `createEventsTableInstance` would match that, and make it clear at the call site which one you're getting.

src/nwb/hdmf/table/MeaningsTable.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  std::vector<DataSpecPtr> specsToUse = columnSpecs;
+  if (specsToUse.empty()) {
+    specsToUse = createDefaultDataSpecs(valueDataType);
+  }
+
+  // Call parent initialize method. This initializes the DynamicTable and
+  // creates the description attribute. Column names are written by
+  // DynamicTable::finalize(). ElementIdentifiers ids are set in the
+  // constructor.
```
rly: `finalize()` doesn't write the column names — `addColumnName` flushes them as columns are added, which the CHANGELOG entry for this PR states explicitly. The same stale claim appears at `DynamicTable.hpp:102` and `ElectrodesTable.cpp:159`.

The last sentence here is also wrong: `ElementIdentifiers` ids aren't set in the constructor, they're written by `setRowIDs`.

CHANGELOG.md
```diff
@@ -16,21 +17,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 * **[BREAKING]** Moved `disableSWMRMode` option from `HDF5IO` constructor to a new `HDF5IO::startRecording(bool disableSWMRMode)` overload. The `BaseIO`-compliant `startRecording()` override is preserved and defaults to SWMR enabled. 
    * **Migration Note**: Code using `HDF5IO(path, true)` must be updated to `HDF5IO(path)` followed by `startRecording(true)`. When the `HDF5IO` object is held as a `std::shared_ptr<BaseIO>` (e.g., from `createIO`), downcast with `std::dynamic_pointer_cast<HDF5IO>` to access the overload. (@oruebel [#297](https://github.com/NeurodataWithoutBorders/aqnwb/pull/297))
+* **[BREAKING]** Refactored `DynamicTable` so it creates and initializes its `ElementIdentifiers` `id` column internally during `initialize()`. As part of this change, `setRowIDs()` now takes only the row ID values and writes them directly to the table's built-in `id` column instead of accepting a separate `ElementIdentifiers` object. In practice this aligns the API with the intended usage, since callers should have always used the table's own `id` dataset. (@oruebel, [#302](https://github.com/NeurodataWithoutBorders/aqnwb/pull/304))
```
rly:
```suggestion
* **[BREAKING]** Refactored `DynamicTable` so it creates and initializes its `ElementIdentifiers` `id` column internally during `initialize()`. As part of this change, `setRowIDs()` now takes only the row ID values and writes them directly to the table's built-in `id` column instead of accepting a separate `ElementIdentifiers` object. In practice this aligns the API with the intended usage, since callers should have always used the table's own `id` dataset. (@oruebel, [#304](https://github.com/NeurodataWithoutBorders/aqnwb/pull/304))
```

oruebel: Fixed in [dd449bb](https://github.com/NeurodataWithoutBorders/aqnwb/pull/320/commits/dd449bbf76844af6e825757eaeefaf1d0a23f85e)

src/nwb/event/EventsTable.hpp
oruebel: Fixed in [6507318](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/65073185c9a648e8d12a739d7208f01ef0f1f25c)

src/nwb/hdmf/table/MeaningsTable.cpp
oruebel: Fixed in [03ce200](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/03ce200c5c57878a7f45de3fa7b3933e3428e4a7)

src/nwb/NWBFile.hpp
oruebel: Fixed in [03ce200](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/03ce200c5c57878a7f45de3fa7b3933e3428e4a7)

src/nwb/hdmf/table/DynamicTable.cpp
oruebel: Good catch. Fixed in [8832f4f](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/8832f4f8486930ce4641ab5a6675fe6524d4502c)

src/nwb/hdmf/table/DynamicTable.hpp
oruebel: Fixed in [5f04cab](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/5f04cabe83bb3e0c16161782c35a9c0475abeb98)

oruebel:
@rly thanks for the additional suggestions. I've addressed all comments.

rly (approved):

rly:
Looks good! Thanks for addressing all the comments. I think this is good to go.

oruebel:
> Looks good! Thanks for addressing all the comments. I think this is good to go.

Thanks for the helpful comments! #325 and #320 are the other two PRs for the 0.4 release.

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -158,6 +680,256 @@ Status DynamicTable::finalize()
...
+{
+  auto ioPtr = getIO();
+  if (!ioPtr) {
+    return Status::Failure;
+  }
+
+  for (const auto& colName : m_colNames) {
+    auto col = readColumn<VectorData>(colName);
+    if (col) {
+      registerColumn(col);
```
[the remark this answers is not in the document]
rly: Tracked in #330, which reproduces this and identifies the same cause: `loadConfiguredColumnsFromFile()` rebuilds from the `colnames` attribute, which by HDMF convention omits the `*_index` companions, so a reopened table registers no `VectorIndex`. Resolving here to keep it in one place.

src/nwb/hdmf/table/VectorIndex.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+  // First, read the indices from this VectorIndex
+  SizeArray startArray = {start};
+  SizeArray countArray =
+      count != AQNWB::Types::SizeTypeNotSet ? SizeArray {count} : SizeArray {};
+  SizeArray strideArray = {stride};
+  SizeArray blockArray = {block};
+
+  std::vector<AQNWB::Types::CellValue> indices = VectorData::readCellValues(
+      startArray, countArray, strideArray, blockArray);
```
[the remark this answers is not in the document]
rly: Tracked in #334, which reproduces the stride case and also covers `block > 1`. Resolving here to keep it in one place.

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -158,6 +680,256 @@ Status DynamicTable::finalize()
...
+{
+  std::vector<int> ids(rowCount);
+  int startId = 0;
+  if (m_rowElementIdentifiers) {
+    auto idData = m_rowElementIdentifiers->recordData();
+    if (idData) {
+      const auto& currentShape = idData->getShape();
+      if (!currentShape.empty()) {
+        startId = static_cast<int>(currentShape.front());
+      }
```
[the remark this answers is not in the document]
rly: Tracked in #5, which asks for uniqueness to be enforced on `ElementIdentifiers`. The mixed explicit and automatic ID collision described here is a concrete case of that. Resolving here to keep it in one place.

rly (dismissed):

src/nwb/hdmf/table/MeaningsTable.hpp
```diff
@@ -0,0 +1,112 @@
...
+ */
+class MeaningsTable : public AQNWB::NWB::DynamicTable
+{
+public:
+  /**
+   * @brief Constructor
+   * @param path Path to the object in the file
+   * @param io IO object for reading/writing
+   */
+  MeaningsTable(const std::string& path, std::shared_ptr<AQNWB::IO::BaseIO> io);
```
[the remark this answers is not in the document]
oruebel: Fixed in [4426f06](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/4426f06061eeefc592ac6ad3379128a92d15a547)

src/nwb/event/TimestampVectorData.hpp
```diff
@@ -0,0 +1,106 @@
...
+    Status initialize(Data& data) const override;
+  };
+
+  /**
+   * @brief Constructor
+   * @param path Path to the object in the file
+   * @param io IO object for reading/writing
+   */
+  TimestampVectorData(const std::string& path,
+                      std::shared_ptr<AQNWB::IO::BaseIO> io);
```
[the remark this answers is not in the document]
oruebel: Fixed in [dc603d6](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/dc603d609d6942d8d14dfeeff9b87dab2230c42d)

src/nwb/event/EventsTable.hpp
```diff
@@ -0,0 +1,159 @@
...
+ */
+class EventsTable : public AQNWB::NWB::DynamicTable
+{
+public:
+  /**
+   * @brief Constructor
+   * @param path Path to the object in the file
+   * @param io IO object for reading/writing
+   */
+  EventsTable(const std::string& path, std::shared_ptr<AQNWB::IO::BaseIO> io);
```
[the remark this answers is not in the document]
oruebel: Fixed in [dc603d6](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/dc603d609d6942d8d14dfeeff9b87dab2230c42d)

src/nwb/event/DurationVectorData.hpp
```diff
@@ -0,0 +1,103 @@
...
+    Status initialize(Data& data) const override;
+  };
+
+  /**
+   * @brief Constructor
+   * @param path Path to the object in the file
+   * @param io IO object for reading/writing
+   */
+  DurationVectorData(const std::string& path,
+                     std::shared_ptr<AQNWB::IO::BaseIO> io);
```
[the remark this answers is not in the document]
oruebel: Fixed in [dc603d6](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/dc603d609d6942d8d14dfeeff9b87dab2230c42d)

src/nwb/hdmf/table/MeaningsTable.cpp
```diff
@@ -0,0 +1,82 @@
...
+MeaningsTable::MeaningsTable(const std::string& path,
+                             std::shared_ptr<AQNWB::IO::BaseIO> io)
+    : DynamicTable(path, io)
+{
+}
+
+Status MeaningsTable::validateDataSpecs(
+    const std::vector<DataSpecPtr>& dataSpecs) const
+{
+  return checkRequiredColumnNames({"id", "value", "meaning"}, dataSpecs);
```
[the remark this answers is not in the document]
oruebel: Fixed in [477895a](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/477895aaa555f2a1d8c1a1cfdf24e485eec5b7fe)

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -49,7 +55,84 @@ Status DynamicTable::initialize(const std::string& description)
...
+
+  clearColumns();
+  Status configureStatus = configureDataObjects(effectiveSpecs);
+  return containerStatus && configureStatus;
+}
+
+Status DynamicTable::validateDataSpecs(
+    const std::vector<DataSpecPtr>& dataSpecs) const
+{
+  return checkRequiredColumnNames({"id"}, dataSpecs);
```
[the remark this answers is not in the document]
oruebel: Fixed in [477895a](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/477895aaa555f2a1d8c1a1cfdf24e485eec5b7fe)

src/nwb/file/ElectrodesTable.cpp
```diff
@@ -38,31 +30,59 @@ ElectrodesTable::ElectrodesTable(const std::string& path,
...
+  // TODO: Add support for creating reference columns with chunking to support
+  // resize/append.
+
+  return specs;
+}
+
+Status ElectrodesTable::validateDataSpecs(
+    const std::vector<DataSpecPtr>& dataSpecs) const
+{
+  return checkRequiredColumnNames({"id", "location", "group_name"}, dataSpecs);
```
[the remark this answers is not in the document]
oruebel: Fixed in [477895a](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/477895aaa555f2a1d8c1a1cfdf24e485eec5b7fe)

src/nwb/event/EventsTable.cpp
```diff
@@ -0,0 +1,118 @@
...
+EventsTable::EventsTable(const std::string& path,
+                         std::shared_ptr<AQNWB::IO::BaseIO> io)
+    : DynamicTable(path, io)
+{
+}
+
+Status EventsTable::validateDataSpecs(
+    const std::vector<DataSpecPtr>& dataSpecs) const
+{
+  return checkRequiredColumnNames({"id", "timestamp"}, dataSpecs);
```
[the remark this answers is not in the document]
oruebel: Fixed in [477895a](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/477895aaa555f2a1d8c1a1cfdf24e485eec5b7fe)

src/nwb/epoch/TimeIntervals.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  return specs;
+}
+
+Status TimeIntervals::validateDataSpecs(
+    const std::vector<DataSpecPtr>& dataSpecs) const
+{
+  return checkRequiredColumnNames({"id", "start_time", "stop_time"}, dataSpecs);
+  // TODO Also add support for checking dtype of named columns defined in the
+  // schema
```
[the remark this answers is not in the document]
oruebel: Fixed in [477895a](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/477895aaa555f2a1d8c1a1cfdf24e485eec5b7fe)

src/nwb/hdmf/table/DynamicTable.cpp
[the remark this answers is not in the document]
oruebel: This is a separate issue https://github.com/NeurodataWithoutBorders/aqnwb/issues/330

src/nwb/NWBFile.cpp
```diff
@@ -88,16 +92,31 @@ Status NWBFile::initialize(const std::string& identifierText,
...
+    initStatus = initStatus && createStatus;
+  }
+
+  // Create subject group and its contents if subject metadata is provided
+  if (subjectSpec.has_value()) {
+    const std::string subjectPath =
+        mergePaths(NWBFile::GENERAL_PATH, "subject");
+    if (!ioPtr->objectExists(subjectPath)) {
+      auto subject = AQNWB::NWB::Subject::create(subjectPath, ioPtr);
+      Status subjectInitStatus = subject->initialize(subjectSpec.value());
```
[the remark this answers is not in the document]
oruebel: Fixed in [6146179](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/6146179bc0b2a71d77ef852177fc89426d72ac0a)

src/nwb/hdmf/table/DynamicTable.cpp
```diff
@@ -101,14 +207,431 @@ Status DynamicTable::setRowIDs(
...
+          uint32_t newIndex =
+              static_cast<uint32_t>(existingTargetElements + newTargetSize);
+          indexBuffers[col.name].push_back(newIndex);
+        }
+      } else if (targetColumns.find(col.name) == targetColumns.end()) {
+        // Regular column (not a target of a VectorIndex)
+        auto it = row.find(col.name);
+        if (it != row.end()) {
+          Status appendStatus =
+              appendCellValueToBuffer(columnBuffers[col.name], it->second);
```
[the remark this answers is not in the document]
oruebel: Fixed in [b77a716](https://github.com/NeurodataWithoutBorders/aqnwb/pull/300/commits/b77a71698d39ff363637e43925cab2ee62f4cef6)

rly (dismissed):

rly (approved):

PR #305, opened 2026-07-05
Updated DynamicTable configuration to support row-based recording

### LINKED ISSUE #303: Support row-wise recording of DynamicTable

oruebel:
Fix #303 

**Problem:** The current implementation of `DynamicTable` is mainly designed for the purpose of metadata tables, e.g., `ElectrodeTable`, where we can usually fill in the whole table before the actual acquisition. With the new `EventTable`, however, we need enhanced support for acquiring the data in the table as part of the acquisition workflow. 

**Goal:** Update `DynamicTable` and `EventTable` to simplify row-based recording and configuration of the table

**Changes:**
- **Support configuration of the columns of a DynamicTable directly via initialize, rather than just incremental via`addColumn`**
    - [X]  Added `Data::DataSpecBase` and `Data::DataSpec<T>` to support runtime configuration of `DynamicTable` column layouts and to decouple table schema definition from object initialization. Each `Data` subclass (e.g., `VectorData`, `ElementIdentifiers`, `TimestampVectorData`, `DurationVectorData`) now exposes a nested `DataSpec` struct that bundles the dataset configuration, description, and any type-specific parameters needed to create and initialize the column. 
    - [X] Added `DynamicTable::createDefaultDataSpecs` static factory to return the default ordered list of `DataSpec` objects for a table (currently just the `id` column). Subclasses (`ElectrodesTable`, `EventsTable`, `MeaningsTable`) override this to append their own required columns, allowing callers to retrieve, inspect, and customize the default column configuration before passing it to `initialize()`.
    - [X] Updated `DynamicTable::initialize`, `ElectrodesTable::initialize`, `EventsTable::initialize`, and `MeaningsTable::initialize` to accept a `std::vector<DataSpecPtr>` column-spec list instead of individual `rowChunkSize` / resolution / flag parameters.
    - [X] Added `DynamicTable::addColumn(DataSpecPtr)` overload consistent with the new `DataSpec`-based API used by `initialize()` and `createDefaultDataSpecs()`.
    - [X] Added `DyanmicTable::validateDataSpecs` (and `DynamicTable::checkRequiredColumnNames` helper function) to validate column specifications as part of `DyanmicTable::initialize` before initialization. Also updated subclassed of `DynamicTable` (e.g., `ElectrodesTable`, `EventsTable`, `MeaningsTable`) to override `validateDataSpecs` to provide their specific validation logic. 
- **Support row-based write of DynamicTable across all columns of the table**
    - [X] Added `DynamicTable::addRow` and `DynamicTable::addRows` methods to support row-based data insertion into a `DynamicTable`. Rows are specified as `DynamicTable::RowData` (an `unordered_map` from column name to `CellValue` variant), enabling type-safe, column-keyed writes without requiring callers to manage per-column buffers directly. Row IDs are auto-generated if not provided. 
    - [X] Added `BaseDataType::BaseDataVariant` scalar variant type and `BaseDataType::createEmptyVectorVariant` helper to support runtime-typed single-cell and buffer operations needed by `addRow`/`addRows`. (@cline, @oruebel, [#305](https://github.com/NeurodataWithoutBorders/aqnwb/pull/305))
- **Update Documentation** 
    - [X] Added tutorial to demonstrate acquisition workflow for DynamicTable. The tutorial uses EventsTable as an example and shows: 1) row-based appending of data and 2)column-based write and addition of full columns to a table, and 3) adding of a `MeaningsTable` to a columns 
    - [X] Updated the CHANGELOG
- **Created Follow-up issues** 
    - [X] Appending rows for reference columns is not yet working. See https://github.com/NeurodataWithoutBorders/aqnwb/issues/313
    - [X] Update `schema_to_aqnwb.py` utility to follow the new structure for `DynamicTable`. See https://github.com/NeurodataWithoutBorders/aqnwb/issues/314


oruebel:
Note, validation test failures right now are likely due to: 1) this branch already updates to hdmf-common schema 1.10 where `MeaningsTable.target` has been changed to an atrribute, and 2) PyNWB has not been released yet with hdmf-common 1.10 so it still expects `MeaningsTable.target` to be a link. This should be fixed once PyNWB has been released.

PR #308, opened 2026-07-13
Add ros3 benchmark demo

oruebel:
- [X] Added simple demo for benchmarking ROS3 performance 
- [X] Updated CI for demo runs to use Ubuntu and add the new ROS3 demo
- [X] Added new `BaseIO::findObject` and `RegisteredType::findOwnedObject` methods to simplify searching for objects by name 
- [X] Added override for `HDF5IO::findObject` to improve search performance
     - [X] Created follow-up items #310 to further optimize search.
- [X] Fixed `install-config.cmake` to propagate HDF5 include directories to downstream consumers of the `aqnwb::aqnwb` target, allowing them to locate HDF5 headers regardless of where HDF5 is installed on the system. 
- [X] Merged changes from #309 which adds remfile support


oruebel:
Completed review

PR #309, opened 2026-07-14
Add remfile VFD for reading remote files over HTTP(S)

bendichter:
Targets #308 (`add_ros3_benchmark_demo`) rather than `main`, since it builds on that PR's `openS3` work and benchmark demo.

## What this adds

An alternative to the ROS3 VFD for remote reads: [remfile-cpp](https://github.com/catalystneuro/remfile-cpp), a C++ port of the Python [remfile](https://github.com/magland/remfile) package, implemented as a read-only HDF5 Virtual File Driver. It serves HDF5's reads from an in-memory chunk cache and fills misses with HTTP range requests via libcurl, with adaptive read-ahead.

- **`HDF5IO::openRemote()`** alongside `openS3()`, guarded by `AQNWB_HAVE_REMFILE_VFD`.
- **CMake option `AQNWB_USE_REMFILE`** (ON by default except on Windows; silently disabled if libcurl is absent). It prefers an installed remfile package and otherwise fetches `catalystneuro/remfile-cpp` at the pinned tag `v0.1.2` via `FetchContent`. `install-config.cmake` gained a `find_package(remfile)` so downstream consumers of an installed aqnwb resolve the target.
- **`ros3_benchmark` demo** takes an optional driver argument (`ros3` | `remfile`), and CI runs the C++ demo with both — completing the matrix with the Python `--force-remfile` runs already in #308.
- Docs (`reads3.dox`), demo README, and CHANGELOG updated.

## How it compares

The `build-demo` CI job on this PR runs the demo against the same DANDI file with both drivers. From that run:

| | ROS3 | remfile |
|---|---|---|
| C++ demo, `read_io` (open) | 0.74 s | **0.63 s** |
| C++ demo, `read_slice` | 1.35 s | **0.14 s** |
| **C++ demo, total** | 4.10 s | **0.77 s** |
| Python (PyNWB), total | 15.99 s | 1.11 s |

I also benchmarked the two drivers head-to-head in a single process on a larger read (30,000 x 384 slice of `ElectricalSeriesAp`, 23 MB), alternating driver order, means of 6 runs each:

| Phase | ROS3 | remfile |
|---|---|---|
| open (`H5Fopen`) | 3.26 s | **0.33 s** |
| metadata | 0.09 s | **0.00 s** |
| bulk read (23 MB) | **1.06 s** | 2.11 s |
| **total** | 4.40 s | **2.44 s** |

The win is concentrated in file open (~10x), where hundreds of small scattered metadata reads are exactly what a read-ahead cache is for. On a single large contiguous read ROS3 is somewhat leaner (remfile speculatively over-fetches ~24% past the request), though the two overlap run to run — the network is noisy enough that I would not lean on the exact ratio.

Two things a benchmark does not show, and which motivated this:

1. **ROS3 requires HDF5 to be built with ROS3 support**, which many common builds lack — Homebrew's HDF5 and every conda `h5py` on my machine do not have it. I had to build a dedicated env just to benchmark ROS3. remfile needs only libcurl.
2. It works with **any HTTP(S) server supporting byte ranges**, not just S3 — including presigned URLs, which reject the HEAD requests some clients rely on (remfile derives file size from a 1-byte range request instead).

## Notes

- The driver is read-only; `write`/`truncate` fail, and `open` rejects anything but `H5F_ACC_RDONLY`.
- remfile-cpp is compatible with HDF5 1.10 through 2.x (the `H5FD_class_t` layout changed in 1.14; both are handled).
- Benchmarking surfaced one real bug in the port, fixed upstream before tagging: the Python smart loader only ever grows its request geometrically, so a large read ramped into 11 sequential range requests, each paying a round trip. Never fetching less than the current read needs brought that to 5 requests and bulk-read throughput to parity with ROS3.
- Verified end-to-end on macOS and in CI on Linux: `FetchContent` of the pinned tag, build, and read from DANDI. Data read matches Python remfile byte for byte. Format-clean under the project's clang-format config.

### Pre-existing CI failures (inherited from #308, not caused by this PR)

- `test (windows-latest)` fails in the *"Install dependencies - windows"* step, before aqnwb is configured at all — #308 fails at the identical step. (remfile also defaults to OFF on Windows.)
- `run-scripts` / *"Test generating classes from schema files"* fails identically on #308; it exercises NWB schema codegen, which this PR does not touch.

Both pass on `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bendichter:
One thing worth flagging for context on the remfile-vs-ROS3 comparison, since #308 added ROS3 support and this PR adds remfile alongside it.

The ROS3 VFD was substantially improved in HDF5's `develop` branch. [HDFGroup/hdf5#6478](https://github.com/HDFGroup/hdf5/commit/1325d30b21ee1328e88103b269a74cad699fa7b6) moved the eager multi-MiB read that ROS3 previously did at file open to the first read instead, and added an internal LRU block cache. The backend also switched from libcurl to aws-c-s3.

I rebuilt HDF5 `develop` with ROS3 and reran the benchmark from the `ros3_benchmark` demo (23 MB `ElectricalSeriesAp` slice from DANDI, alternating drivers, medians of 8 runs):

| Phase | ROS3 (HDF5 2.1) | ROS3 (develop) | remfile |
|---|---|---|---|
| open (`H5Fopen`) | 3.26 s | 1.51 s | 0.35 s |
| read (23 MB) | 1.06 s | 1.58 s | 1.20 s |
| total | 4.40 s | 3.08 s | 1.56 s |

ROS3's open time roughly halved. remfile is still faster to open (about 4x now, down from ~10x), and end to end on this workload, but the gap is narrower than it was and will keep narrowing as that block cache matures. The read times are close enough that I would not read much into the bulk-read comparison.

Nothing here needs to change in the PR. I mention it mainly so that whoever documents the two drivers for users has the current picture, and because once these HDF5 changes are in a release it may be worth rerunning the demo to refresh any numbers we cite. The [HDF Group discussion](https://github.com/HDFGroup/hdf5/discussions/6527) has the details.

docs/pages/userdocs/reads3.dox
```diff
@@ -34,6 +34,30 @@
...
+ *
+ * \code
+ * auto io = std::make_shared<AQNWB::IO::HDF5::HDF5IO>(url);
+ * io->openRemote();
+ * \endcode
+ *
+ * Compared to ROS3, the remfile VFD:
+ * - does not require HDF5 to be built with ROS3 support (it only requires libcurl,
+ *   and is enabled with the ``AQNWB_USE_REMFILE`` CMake option, ON by default on
+ *   non-Windows platforms)
```
oruebel:
```suggestion
 * - does not require HDF5 to be built with ROS3 support. It only requires libcurl and the light-weight remfile-cpp. remfile support 
 *   is enabled with the ``AQNWB_USE_REMFILE`` CMake option (OFF by default). If remfile-cpp is not installed, then AqNWB will build it directly 
 *   as part of its built process, so only libcurl is required as a separate install.
```

docs/pages/userdocs/reads3.dox
```diff
@@ -34,6 +34,30 @@
...
+ *
+ * As an alternative to ROS3, AqNWB can read remote files with the
+ * [remfile-cpp](https://github.com/catalystneuro/remfile-cpp) virtual file driver, a C++
+ * port of the Python [remfile](https://github.com/magland/remfile) package. Instead of \ref AQNWB::IO::HDF5::HDF5IO::openS3 "openS3",
+ * call \ref AQNWB::IO::HDF5::HDF5IO::openRemote "openRemote":
+ *
+ * \code
+ * auto io = std::make_shared<AQNWB::IO::HDF5::HDF5IO>(url);
+ * io->openRemote();
+ * \endcode
```
oruebel: Would be useful to add a `tests/examples/test_ros3_read_example.cpp` example akin to "https://github.com/NeurodataWithoutBorders/aqnwb/blob/add_remfile_vfd/tests/examples/test_ros3_read_example.cpp" so that we can include the code examples as a tested `\snippet`

CMakeLists.txt
```diff
@@ -106,6 +106,47 @@ set_target_properties(aqnwb_aqnwb PROPERTIES
...
+# ---- Optional remfile VFD for reading remote files over HTTP(S) ----
+# remfile provides an HDF5 virtual file driver that reads remote files with
+# HTTP range requests and adaptive read-ahead caching. Unlike the ROS3 VFD,
+# it does not require HDF5 to be built with ROS3 support.
+if(WIN32)
+  set(aqnwb_use_remfile_default OFF)
+else()
+  set(aqnwb_use_remfile_default ON)
+endif()
+option(AQNWB_USE_REMFILE "Enable reading remote files over HTTP(S) via the remfile VFD (requires libcurl)" ${aqnwb_use_remfile_default})
```
oruebel:
```suggestion
set(aqnwb_use_remfile_default OFF)
option(AQNWB_USE_REMFILE "Enable reading remote files over HTTP(S) via the remfile VFD (requires libcurl)" ${aqnwb_use_remfile_default})
```

In order to keep default dependencies of AqNWB light and since not all acquisition system will need streaming support, I propose to have remfile (and ros3) disabled by default. The only reason we don't need an explicit (enable/disable) option for ROS3 in the AqNWB built is because this part of the HDF5 built so we can infer whether to enable/disable ROS3 based on if HDF5 was built with ROS3 support or not.

docs/pages/userdocs/reads3.dox
oruebel: Note, examples are run as part of the unit tests. I.e., we should also update the `.github/workflows/tests.yml` then to make sure `libcurl` is being installed and `AQNWB_USE_REMFILE` is being enabled in the built to make sure it is being tested in the CI.

oruebel (commented):
@bendichter overall the code looks great! Thanks for this great contribution! Just a few simple things to add tests and update the default build behavior. Otherwise this looks great.

CMakeLists.txt
oruebel: Fixed in [cf474cc](https://github.com/NeurodataWithoutBorders/aqnwb/pull/309/commits/cf474cc0ca436c6642c31092c21cb050635c0537)

docs/pages/userdocs/reads3.dox
oruebel: Fixed in [92e5eff](https://github.com/NeurodataWithoutBorders/aqnwb/pull/309/commits/92e5eff99c51e862759976362cf65c8b35439d76) and [b37ee65](https://github.com/NeurodataWithoutBorders/aqnwb/pull/309/commits/b37ee6578574ec168f1f85f1c016853c63d75307)

oruebel (approved):
I added the unit test and updated build and docs. Looks good.

